PR #1, opened 2019-02-01
moved the libraries in ILL/Utils ILL/Vec in Shared/Libs/DLW

DmxLarchey:
and added most of the external libraries for the proof of H10
to Shared/Libs/DLW as well


DmxLarchey:
Yannick, can you just check it compiles still fine for you ... I just moved files around
and completed  the libraries. On my computer everything still compiles

yforster:
Yep, everything works!

PR #2, opened 2019-02-02
Transfer of n-ary DPRM + reduction from TM_Halt to H10

DmxLarchey:
Hi Yannick,

Could you try to compile ? And after that, maybe review the structural of the file to see if you find it logical and also good fit with the paper ...


yforster:
Compiling works! And also I think the structure is good. The coqdoc has to be improved though, but that hold's for the whole development. So I think we can merge

DmxLarchey:
Ok I merge

PR #3, opened 2019-02-06
Mm fractran, simplified backward step proof

DmxLarchey:
Hi Yannick ... I propose a shortened proof of the backward simulation one step FRACTRAN -> one step MM. The trick is the one with functionality of FRACTRAN and totality of MM combined with one step forward. 

Also some improvements of the code for better corresp. with the paper.

Do you agree merging in wip ?

### DIFF
--- a/H10/DPRM.v
+++ b/H10/DPRM.v
@@ -37,12 +37,11 @@ Section DPRM_n.
 
   (* There is a FRACTRAN program simulating R *)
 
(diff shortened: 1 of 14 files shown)


DmxLarchey:
Hi Yannick,

I completely removed the need for Fractran programs to be regular ... in fact, I changed the proof that R* is diophantine and it does not need boundedness anymore ... I am going to update the paper as well in a separate branch ... but I strongly suggest we adopt the change because regularity is really not needed anymore

PR #6, opened 2019-05-02
New repo structure, added Turing machines and lambda-calculus

yforster:
This is a draft PR and not ready for merging.

It contains / should contain

- [x]  Reorganisation of file structure. Our files go to `theories`, external libraries go to `external`. Also all names are absolute now (by using `-Q` in `_CoqProject` instead of `-R`). Considering that we're well over 10000 lines by now, it's really hard to force files to have distinct names and it's easier to just force fully qualified names when importing.
- [x] Turing machines from http://github.com/uds-psl/CoqTM/. Including a reduction from abstract stack machines to multi-tape TM halting and a reduction from multi-tape to single-tape TM halting. Also includes a universal Turing machine. Directory `TM`
- [x] Lambda calculus development with extraction framework from https://github.com/uds-psl/certifying-extraction-with-time-bounds. Including a reduction from multi-tape TM halting to L halting and a reduction of `H10c` to L halting. Directior `L`
- [x] relevant parts of the stack machines from https://www.ps.uni-saarland.de/extras/cbvlcm2/. This doesn't compile yet, probably because I mixed up definitions from different subprojects. @fakusb will have a look. This will include a reduction from L halting to abstract stack machines (closing one possible loop between L, TMs, MM, PCP and H10c). Directory `LAM`
- [x] proper documentation, relevant files in `Problems` and `Reductions`
- [x] adapted README and credits for authors (Fabian Kunze and Maximilian Wuttke)
- [x] the website directory has to be moved
- [x] We need build instructions. Parts of this depend on Equations. I might be able to rework those parts, but Equations is easy to install using `opam` and shipped with the Coq package, so I'm not too worried.
- [x] Reduce multi-tape TM halting to single-tape TM halting
- [ ] Disable warnings
- [ ] Prefix every `Print Assumptions` with a `Check` to make output readable


fakusb:
TODO:
- [ ] use TM-def in PCP-reduction?

yforster:
> TODO:
> 
>     * [ ]  use new TM-def in PCP-reduction?

As discussed personally, we won't use the new definition of TMs in the PCP reductions. `PCP/singleTM.v` contains a reduction from the new TM definition to the "old" definition, which factors the reduction and avoids reworking the huge proof scripts in `PCP/TM_to_SRH.v`.

fakusb:
Very reasonable, sometimes a second, more specialized definition with translation lemmas is the way to go.

yforster:
Alright, this is ready for review and discussion now. I've merged in the MM2 commits from `wip` already.

I think that I merged in Turingmachines and lambda-calculus is fairly uncontroversial, but

- the reorganisation in `theories` and `external` and
- the move to absolute naming for imports

could be more controversial. @DmxLarchey  what do you think?

DmxLarchey:
Hi Yannick, Fabian,

How do I get the code from `external` (apparantly a kind of git specific symlink) to be pulled by git ... for now the directories in `external` are empty in my local copy of  the `extraction` branch ?

Best,

D.

yforster:
I've updated the build instructions off the top of my head, but there was something missing, sorry. The following works for me:

``` sh
git clone https://github.com/uds-psl/coq-library-undecidability.git dlib
cd dlib
git checkout exraction
git submodule init
git submodule update
make -j 8```

**Comment by DmxLarchey** [comment]
Acting as a newbie ... I am running into problems compiling template Coq ... further instructions for that ?

**Comment by DmxLarchey** [comment]
Btw template coq seems to require 8.8 whereas MetaCoq is for 8.9 ?

**Comment by DmxLarchey** [comment]
> Btw template coq seems to require 8.8 whereas MetaCoq is for 8.9 ?

I get 

```COQC theories/kernel/uGraph.v
CAMLOPT -c -for-pack Template_coq src/quoter.ml
File "src/quoter.ml", line 19, characters 70-95:
Error: Unbound type constructor Context.Rel.Declaration.t
Makefile.coq:610 : la recette pour la cible « src/quoter.cmx » a échouée
make[5]: *** [src/quoter.cmx] Erreur 2
make[5]: *** Attente des tâches non terminées....
File "./theories/kernel/uGraph.v", line 80, characters 66-71:
Error: No interpretation for string "nil".
```

while trying to compile template-coq ...

yforster:
Template Coq (and MetaCoq) only work under 8.8 at the moment (the documentation might be broken and claim other things). I hope that this changes in the next couple of weeks though. For now, this means that we can also only compile under 8.8. I'll try to push for a move to 8.9.

The above instructions work for me with Coq version
```
The Coq Proof Assistant, version 8.8.2 (March 2019)
compiled with OCaml 4.07.0
```
(now they do, the `_Coqproject` file was missing before)

DmxLarchey:
Works better with 8.8.1 indeed ... Btw we should document why template Coq is used because it involves axioms;-)

DmxLarchey:
Ouch ... compiling template-coq under 8.8.1 fails after a (long) while ... which one do you have ? 8.8.2 ?

yforster:
They don't leak through to any of our theorems though, right? The reduction from mTM to L is still axiom-free for instance.

And yes, 8.8.2.

DmxLarchey:
What can we compile w/o unstable external tools ?

DmxLarchey:
Got to go ... I will try 8.8.2 later on

yforster:
I think once we've figured out the correct build instructions it's not unstable anymore, because they submodules are fixed to specific commits and don't change.

On the mid-term, the external tools (well, at least MetaCoq) will be stable, I hope.

DmxLarchey:
I tried with 8.8.2 and still stuck in compiling Template-Coq independently of the rest ...

```
Moving  *.ml* to src/eAst.ml src/eAst.ml.d src/eAst.mli src/eAst.mli.d src/eAstUtils.ml src/eAstUtils.ml.d src/eAstUtils.mli src/eAstUtils.mli.d src/eInduction.ml src/eInduction.ml.d src/eInduction.mli src/eInduction.mli.d src/eLiftSubst.ml src/eLiftSubst.ml.d src/eLiftSubst.mli src/eLiftSubst.mli.d src/eTyping.ml src/eTyping.ml.d src/eTyping.mli src/eTyping.mli.d src/extract.ml src/extract.ml.d src/extract.mli src/extract.mli.d src/template_coq_extraction_plugin.mlpack src/template_coq_extraction_plugin.mlpack.d
mv: la cible 'src/template_coq_extraction_plugin.mlpack.d' n'est pas un répertoire
```
which blocks the compilation afterwards .... seems there is a problem in the building scripts for TC ...

Yannick could you please try to recompile TC from scratch with 8.8.2 ?

yforster:
I can recompile from scratch and it also works on Dominik's machine. 

`8.8.2` is essential for now, but the TemplateCoq beta for `8.9` is expected for the end of the week.

Can you give it another go after a fresh clone? Just to exclude the chance that there are stray files initially compiled with `8.9` that now block compilation which are not removed by `make clean`.

DmxLarchey:
I did recompile from a really fresh install:

```
opam switch for8.8.2 --alias-of system
eval `opam config env`
opam install coq.8.8.2
opam install coqide.8.8.2
opam install coq-equations.1.1+8.8
```

then

```
git clone git@github.com:uds-psl/coq-library-undecidability.git
cd coq-library-undecidability
git checkout extraction
git pull
git submodule init
git submodule update
make -j 16 all
```
and it still fails in the compilation of `TemplateCoq`. I noticed the following warning

```
Warning: no common logical root
Warning: in such case INSTALLDEFAULTROOT must be defined
Warning: the install-doc target is going to install files
Warning: in orphan_TemplateChecker_Template_TemplateChecker
```

and shortly after it fails. What is `INSTALLDEFAULTROOT` ?

yforster:
That's really really strange. I'm using opam2, thus I have to replace the first two command by

```
opam switch create for8.8.2 4.07.1+fp+flambda
eval $(opam env)
```

Using the exact same commands afterwards it compiles. Can you send me a dump of the whole output of make by mail?

It could be opam v1 which is the problem, but I don't think so.

The `INSTALLDEFAULTROOT` is only relevant if we're using `make install` I think. I'm getting the same warning and can compile anyways, so I would guess that's unconnected to the problem, but not sure.

DmxLarchey:
I did run the following under `/bin/tcsh` for `opam 1` and `Coq 8.8.2`

```
git clone git@github.com:uds-psl/coq-library-undecidability.git
cd coq-library-undecidability
git checkout extraction
git pull
git submodule init
git submodule update
cd external/template-coq/
make all |& tee make.out
```
The result in [make.out on PasteBin](https://pastebin.com/50rfVEgv)

DmxLarchey:
Ok I finally nailed it ... **the `OCaml version 4.02.3` is the faulty one** ... I did reinstall everything with the following, so under `OCaml version 4.07.0` and now it compiles; with warnings but it finishes ...

```
opam switch for8.8.1 --alias-of 4.07.0
eval `opam config env`
opam pin add coq 8.8.1
opam install coqide
opam install coq-equations

git clone git@github.com:uds-psl/coq-library-undecidability.git
cd coq-library-undecidability
git checkout extraction
git pull
git submodule init
git submodule update
make -j 16 all
```
I strongly suggest we stress which version of `opam`, `OCaml`, `coq` and `coq-equations` should be used in the `README` otherwise user could get discouraged very quickly ....

Why is the compilation of `./LAM/TM/StepTM.v` taking so long and generates so many warnings all over the place ?

I also suggest that the output of `make` makes sense ... so not so many warnings, and also, `info` messages like `Closed under the global context` should be preceded with a `Check <something>` so that we know what is closed ...

`L/Computability/Rice.v` is taking ages as well ... to much automation there? More to come ...

yforster:
> Ok I finally nailed it ... **the `OCaml version 4.02.3` is the faulty one** ... I did reinstall everything with the following, so under `OCaml version 4.07.0` and now it compiles; with warnings but it finishes ...

Ohh, perfect. I didn't think about that at all.

> ```
> opam switch for8.8.1 --alias-of 4.07.0
> eval `opam config env`
> opam pin add coq 8.8.1
> opam install coqide
> opam install coq-equations
> 
> git clone git@github.com:uds-psl/coq-library-undecidability.git
> cd coq-library-undecidability
> git checkout extraction
> git pull
> git submodule init
> git submodule update
> make -j 16 all
> ```

We could simply copy-paste this to the README. Once we're semi-stable creating an opam package will resolve such issues.

> Why is the compilation of `./LAM/TM/StepTM.v` taking so long and generates so many warnings all over the place ?

It's taking long because it's a huge file verifying the core parts of the `L -> TM` reduction. There's lots of automation happening.

> I also suggest that the output of `make` makes sense ... so not so many warnings, and also, `info` messages like `Closed under the global context` should be preceded with a `Check <something>` so that we know what is closed ...

We should disable many warnings, I agree. And I also agree about the `Closed under global context` things (I've added it to the TODO list in the first post).

> `L/Computability/Rice.v` is taking ages as well ... to much automation there? More to come ...

That's fairly quick for me. @fakusb can you run it and see whether it's slow for you as well?

theories/L/Tactics/Extract.v
```diff
@@ -417,7 +417,6 @@ Fixpoint extract (env : nat -> nat) (s : Ast.term) (fuel : nat) : TemplateMonad
     n <- (tmEval cbv (String.append (name_of s) "_term") >>= tmFreshName) ;;
     i <- tmTryInfer n (Some cbn) (extracted a') ;; 
       ret (@int_ext _ _ i)
-  | tMeta _ =>     tmFail "tMeta is not supported"
```
DmxLarchey: I now get compilation failure because `Non exhaustive pattern-matching: no clause found for pattern tMeta _` ...

README.md
```diff
@@ -27,14 +27,32 @@ Feel free to contribute or start using the problems!
...
 ## How to build
 
-- The subprojects are currently in subdirectories in `theories`, roughly corresponding to papers or theses covering them.
-- External libraries are in the `external` directory
-- `git submodule init && git submodule update` initialises the external libraries
-- You need to install the [Equations](https://mattam82.github.io/Coq-Equations/) package for Coq, see the page for detailed instructions
-- `make all` builds all subprojects by calling `make all` of the corresponding subproject's makefile
+### Required packages
+
+You need `Coq 8.8.1` or `8.8.2` built on OCAML `>= 4.02.3` and the [Equations](https://mattam82.github.io/Coq-Equations/) package for Coq. If you're using opam 2 you can use the following commands to install the dependencies on a new switch:
```
DmxLarchey: **`OCaml 4.02.3` did not work for me**. I only tested `OCaml 4.07.0` successfully

yforster: You're right, fixed (also in the `opam` file)

theories/L/Tactics/Extract.v
yforster: I changed the `template-coq` version to the newest version in the progress of tracking down the build error. You need to run `git submodule update` and then `make deps` to update Template Coq. I've untangled building the dependencies and building our library in the Makefile, but I'm not convinced that's good. 

I don't think we'll change the versions of the libraries often, so a `git submodule update` shouldn't be required often. It's visible if you do `git status` I think.

Once we're rolling out via opam it's a definite non-issue, because opam will take care of such things.

DmxLarchey: Now `TemplateCoq` does not build properly anymore ... because older `.vo` are not removed even with `make mrproper` ....

Yep I think `opam` is definitely the way to go for smooth install with these dependencies ... what I experienced is definitely going to scare 99% of our potential users ....

yforster:
You're the beta tester here, sorry... Do a `git clean -xf` inside the 
`template-coq` folder. This will remove all files not tracked by git. Make 
sure that you're in the right folder though, ugly things can happen otherwise.

DmxLarchey:
Look more like alpha testing to me -;)

DmxLarchey:
Still getting

```
Error:
Cannot find a physical path bound to logical path matching suffix
Equations.Prop.
```
after a `git clean -xf`. I will now try a full reload ...

yforster:
Wait a moment, let me check first that this is not a fault on my side

DmxLarchey:
This occurs while compiling the updated `TemplateCoq` only

DmxLarchey:
Anyway even a full reload of the whole project leads me to the same error.

yforster:
There is a `coq-equations.1.2+8.8` package (which is the newest version of equations), I think that's the problem here. So `opam install coq-equations.1.2+8.8` should fix it.

DmxLarchey:
Ok I did not see that one ... `opam update` but still no `coq-equations.1.2+8.8` in the list with `opam 1.2.2` ...

yforster:
It's in the `released` repo:
```
opam repo add coq-released https://coq.inria.fr/opam/released
```

DmxLarchey:
If you use `coq-equations.1.2+8.8` may be I exists only on `opam 2` ... which is going to be kind of a problem because I don't think you can use both opams simultaneously ...

DmxLarchey:
Already had the `released` repo in

```
zingaro template-coq % opam show coq-equations
             package: coq-equations
             version: 1.2~beta+8.8
          repository: coq-released
        upstream-url: https://github.com/mattam82/Coq-Equations/archive/v1.2-beta-8.8.tar.gz
       upstream-kind: http
   upstream-checksum: 1cc8f9edbc55ac189dfd1471a77c3863
            homepage: https://mattam82.github.io/Coq-Equations
         bug-reports: https://github.com/mattam82/Coq-Equations/issues
            dev-repo: https://github.com/mattam82/Coq-Equations.git
              author: Matthieu Sozeau <matthieu.sozeau@inria.fr>, Cyprien Mangin <cyprien.mangin@m4x.org>
             license: LGPL 2.1
             depends: coq (>= 8.8 & < 8.9)
  installed-versions: 1.2~beta+8.8 [for8.8.2 for8.8.1], 1.2~beta+8.9 [for8.9]
  available-versions: 0.9~beta2, 1.0~beta, 1.0~beta2, 1.0~beta2+8.7, 1.0, 1.0+8.7, 1.0+8.8, 1.1+8.8, 1.2~beta2+8.8, 1.2~beta2+8.9, 1.2~beta+8.8, 1.2~beta+8.9
         description: A function definition package for Coq
```

yforster:
You're right, it's an `opam 2` vs `opam 1` problem. This is worse than alpha I'd say...

I've opened a PR for MetaCoq: https://github.com/MetaCoq/metacoq/pull/144

I think if you want to progress for the moment you either have to upgrade to `opam 2` (which is necessary to obtain future versions of Coq I think) or install Equations by hand by downloading [this tarball](https://github.com/mattam82/Coq-Equations/archive/v1.2-8.8.tar.gz) and then running 

```
coq_makefile -f _CoqProject -o Makefile
make -j 8
make install
```

yforster:
You have `1.2~beta2+8.8` in the list, did you try that one? It seems to work with the MetaCoq Travis CR.

DmxLarchey:
Already tried `1.2~beta2+8.8` and it failed with the MetaCoq/TemplateCoq which is linked into our repo.

DmxLarchey:
I moved to `opam 2` on my laptop with `Ubuntu LTS 16.04` ... what a mess with `bubblewrap` ... had to compile it from the GitHub repo.

But then, I still could not find `coq-equations.1.2+8.8`. Realized that I had to update the Coq repo. which pointed to 1.2 ... the upgrade path from opam 1 to opam 2 is really bad ... now I will try to compile TC ....

DmxLarchey:
Ok TC compiles now .... I tried to recompile the whole project but for some reason I do not understand, I had to compile external libraries first before able to compile the rest of the project ...

It seems to me the two other external do not pose so much version issues ? Is there a way to isolate the part of the project that requires TC+Equations in a separate branch ?

DmxLarchey:
Summary of my experience in compiling the project. The `Template Coq` (or `TC`) inclusion puts many restrictions upon which version of `Coq` and `Equations` can be used, and incidentally on which version of `opam` can be used:

* `TC` does not compile with `Coq 8.9`
*  `TC` needs `Equations 1.2` which is only available for `Coq 8.8.*` under `opam 2`
* To install `opam 2` correctly, I needed to install `bubblewrap` for sandboxing; Older version of `bubblewrap` render `cc1` inaccessible; I finally had to disable sandboxing altogether on Fedora 24.
* Even with `opam 2` installed, one can only compile `TC` with `Coq 8.8.2` and `Equations 1.2`

So requirements are very strict to compile `TC`: 

* Switch to [`opam 2`](https://opam.ocaml.org/blog/opam-2-0-0/);
* Maybe disable sandboxing or install a recent version of [`bubblewrap`](https://github.com/projectatomic/bubblewrap); 
* Via `opam 2`, install `coq.8.8.2` and `coq-equations.1.2+8.8`;
* as explained before, download the undecidability library and its dependencies via `GH`;
* **Compile the `external` tools first**, i.e. `smpl`, `template-coq` and `uds-psl-base`;
* Only then one can compile the undecidability library.

I am convinced that these are way too high constraints for a smooth learning curve. Given that I think `TC` is the main source of these tight dependencies, I suggest isolating the part that depends on it in a separate branch for the moment. I suppose this concerns the part dealing with extraction.

DmxLarchey:
Hi all,

This `extraction` branch seems to look as the next `wip` branch but the problem is that the whole structure has changed, all dirs moved to `theories`. Hence it is not wise to fork from `wip` any more. However, I would like to push some new stuff like `µ-recursives` in the public repo and I am stuck because `extraction` seems to be stalling forever...

Is there a kind of deadline for merging the `extraction` PR anytime soon ?

D.

yforster:
I've renamed the branch to `master` and will take it as the new default branch. We're keeping `wip` as `noopam` branch, so this PR doesn't have to be merged.

PR #8, opened 2019-05-06
added reduction from L to TM

fakusb:
I moved the implementation of abstract machines in terms of Turing machines to the LAM-folder, added definitions of the halting problem for L and TMs and put the reduction from HaltL to HaltTM in LAM/L_to_TM.v .
This extends #6.


theories/Problems/TM.v
```diff
@@ -1,3 +1,9 @@
-From Undecidability.PCP Require Export singleTM.
+From Undecidability.TM Require Export TM.
 
-About halt. (* Halting problem for Turing machines *)
+Definition HaltsTM {sig: finType} {n: nat} (M : mTM sig n) (t : tapes sig n) :=
+  exists outc k, loopM (initc M t) k = Some outc.
+
+Definition HaltTM n (S: {sig:finType & mTM sig n & tapes sig n}) :=
```
fakusb: I defined this to be parametric in the number of tapes, but the alphabet is in the tuple itself. Is this ok?

PR #9, opened 2019-06-27
Merged public changes for private complexity development

fakusb:
This commit includes several improvements for L, the extraction framework and the extracted methods:
-improved functions for binary numbers
-improved abstract machines for L (with extraction)
-more extractions with time bounds
-better support for transforming non-tail-recursive functions to tail-recursive functions using a explicit recursion stack. This approach makes verifying space bounds easier, once implemented.


yforster:
@fakusb, what's the status of this? I'm happy to merge soon if it's still relevant, but there are several conflicts at the moment which would have to be resolved first. (It probably makes sense to wait with resolving until the other PRs are merged though).

fakusb:
I investigates this again: 
One Problem with porting my changes to the public repo is that we either need to move the L-extraction to the Metacoq-Extractable monad (no pun  intended) or need https://github.com/MetaCoq/metacoq/pull/329 upstream.

@yforster : Im would like to tackle syncing our repositories when your are back.

PR #10, opened 2019-10-10
workaround for coq 8.10 (windows) compilation

mrhaandi:
1) several "Local" qualifiers were illegal
2) in Kripke obligations were unsolved
3) induction on (with one and two parameters) was not working


mrhaandi:
Coq 8.10.1 fixes the uconstr bug, so this pull request is partially obsolete.

PR #19, opened 2019-10-25
Import new version of the Turing framework (with Hoare)

mwuttke97:
The new version includes the new *Hoare verification framework*. This
framework aims to make verification of Turing machines easier and
faster.

Misc:
- Remove Haskell extractions
- Rename `isRight` into `isVoid`
- Updated alphabet-lift relation that works better with `cbn`.
  + We don't need `vector_tl_nth` any more.
- Various small fixes


mwuttke97:
At least for me, everything compiled just fine.

mwuttke97:
Rebased in #51.

PR #22, opened 2019-10-29
Construction of an H10C-universal µ-recursive function

DmxLarchey:
In `Problems/RA_UNIV.v`, we build a universal µ-recusive algorithm  `ra_univ : recalg 1`. Its size is 8708.

`ra_univ` is µ-recursive algorithm with 1 parameter which is **universal** w.r.t. elementary Diophantine constraints `h10c`, as established in `Reductions/H10C_RA_UNIV.v`.

To put it simple, its termination predicate `RA_UNIV_HALT` can simulate the solvability of any list of elementary Diophantine constraints. Hence **the termination predicate of this particular µ-recursive algorithm is undecidable**.

The reason why I choose to simulate H10C (and not H10, MM, etc) is that lists of constraints allow for a much easier decoding using µ-recursive algorithms (via Gödel Beta) whereas decoding other structures like eg. `recalg n` or H10 require traversal of trees which is quite hard to implement using µ-recursive schemes. You need to simulate _course-of-value recursion_ which is hard. I suppose `L` can do that but I am not familiar enough with this part of the project.

Hope this is enough to derive universal machines for other languages like MM, FRACTRAN, etc.


DmxLarchey:
In `Reduction`, I need to rework the structure of `MM_to_FRACTRAN.v` and `MM_to_MMA2.v` to factor out `FRACTRAN_REG_HALTING`.

DmxLarchey:
This PR should be clean now. If Travis works out, I request merge into `master`. It should not be too intrusive over the `master` branch: most changes in there are just cleanups, and the rest of the contributions occurs in new files.

DmxLarchey:
The conflict against `master` comes from the Trakthenbrot contribution in _CoqProject ... I will look if I can upload this commit into this PR branch ...

DmxLarchey:
Resynced with `master` with `git merge` ... no pending conflicts anymore

DmxLarchey:
Some remarks for further improvements of the whole project, to keep them in mind.

The parts of the project in `TM/*`, `LAM/*` and `L/*` generate way too much warnings or unintelligible output during the build process.

The following files are taking too long to compile, stalling the whole build process.

- `MuRec/ra_mm_env.v`
- `LAM/TM/StepTM.v`
- `L/TM/TMEncoding.v`

I also suggest on big Reductions file in which we collect all the reductions that are actually needed to check for the completion of the graph of reductions we advertise in e.g. papers, talks, etc.

yforster:
Very happy to merge this as well. I agree that we have to clean up the files and at least give instructions how to build without running in a 50 minutes compilation. It's on my TODO list which also includes reworking the directory structure as suggested somewhere else here.

No reason to not merge atm I think though, right?

DmxLarchey:
You can merge this one !

DmxLarchey:
Did merge `master` in to stay in sync ... I hope CI will work. It did compile on mine.

DmxLarchey:
Seems ok for merge now !!

PR #23, opened 2019-11-06
H10UC + FMsetC + HSC + LPolyNC

mrhaandi:
Contains 
- undecidability of uniform Diophantine constraint solvability (`H10UC`) by reduction from elementary Diophantine constraint solvability (`H10C`)
- undecidability of finite multiset constraint solvability (`FMsetC`) by reduction from uniform Diophantine constraint solvability (`H10UC`)
- undecidability of recognizing axiomatizations of Hilbert-style calculi (`HSC_AX`) by reduction from binary modified Post correspondence problem (`BMPCP`)
- undecidability of provability in a fixed Hilbert-style calculus (`HSC_PRV ΓPCP`) by reduction from binary modified Post correspondence problem (`BMPCP`)
- undecidability of binary modified Post correspondence problem (`BMPCP`) by reduction from modified Post correspondence problem (`MPCP`)
- undecidability of linear polynomial (over ℕ) constraint solvability (`LPolyNC`) by reduction from finite multiset constraint solvability (`FMsetC`)
- notion of Turing-reducibility for later use in undecidability of system F inhabitation
- coq 8.10.1 compatibility
- less `Require Export Shared.Prelim` to prevent clashes with ssreflect

Requires coq 8.10.1


yforster:
I can confirm this compiles (on my machine) under `Coq 8.10.1`. 

I can add Travis tests for `8.10.1`, but we have to decide how to set up the library from here on (and we might as well take this PR to discuss).

One way would be to have a `minimal` branch (which would be the current `noopam` branch), which is allowed to depend on very standard libraries (like `mathcomp` and `equations`), but nothing else. This way, we ensure compilation on limited systems like Windows. Since there are no dependencies, we can have the rule that the `minimal` branch compiles under the current Coq version, and potentially the current Coq version only.

The downside is that we diverge from the more feature-rich `master` branch. Here we have to rely on the external dependencies (in this case that's really only the `coq-metacoq-template` package) quickly catching up with the current Coq version, which is not too realistic.

So we could go to a setting where we have 

- `minimal`, containing everything we have on `noopam` atm + Andrej's new work, compiling under `8.10.1` only
- `master`, containing everything we have on `master` atm, compiling under `8.8.x` and `8.9.x`

@DmxLarchey what do you think?

DmxLarchey:
Do we try to ensure that `master` is a superset of that branch, ie to we try to merge it into master ?

yforster:
As much as possible I would say. But since Andrejs code *only* compiles on 8.10 and master *only* compiles on 8.8 and 8.9, some bits of the code can only be ported after all dependencies of master are ported. In the meantime, master and minimal will be incomparable, they both will have reductions the other one hasn't.

mrhaandi:
I would argue that the `minimal` branch should depend on at most the current coq version (and not even libraries such as `mathcomp` and `equations`). The reason is twofold.
First, it would be easier to ensure long term maintainability (maybe even without unjustified warnings). For example, `mathcomp` in each iteration renames quite a few methods keeps breaking my code.
Second, it could be the to-go place to get in touch with the library, without being a type theoretician, or even familiar with coq, and without worrying that some dependency (in a few years) may start to make trouble on some OS.
Also, along the lines of getting in touch with the library, a small, clean, and well presented subset would be good. I remember my first time with the library where it took be a while to find the definition of TM (somewhere in the middle of PCP line 163).
Seeds definitely need to be part of `minimal`, so other people can effortlessly use a seed.

PR #26, opened 2019-12-10
A fork for smoother reduction dio_form -> dio_elem

DmxLarchey:
Hi Yannick,

This is in preparation for the journal paper.

* I did reimplement `dio_form` so that atoms are just `x ≐ ⌞n⌟`, `x ≐ y`, `x ≐ y ⨢ z` and `x ≐ y ⨰ z` which allows for a much smoother reduction to `dio_elem`, both in the code which is now much nicer, and most importantly in the explanation part that is going to be included in the journal paper. 

* Also, I did strengthen the notion of Diophantine polynomial to that of _Diophantine function_ encoded as `dio_fun` and denoted `𝔻F`. Now we can work with arbitrary functions in Diophantine shapes, provided there is a Diophantine encoding of their graph. This allows for less cumbersome definitions.

* Finally a **major** cleanup of automation that is now very satisfactory ... basically, the new `dio auto` tactic fails nearly _nowhere_. 

* I added a proof of the full statement of _Luca's theorem_ and a proof of _Lagrange's theorem_ (stating positive integers are sums of four squares).

* And yes, I tested _forking_.

### DIFF
--- a/theories/H10/ArithLibs/Zp.v
+++ b/theories/H10/ArithLibs/Zp.v
@@ -702,6 +702,60 @@ Section Zp.
       rewrite Z2Zp_plus; f_equal.
       apply Z2Zp_opp.
     Qed.
+   
+    Section Z2Zp_canon.
+
+      (** Find a representative in the A interval [0,Z.of_nat p[ *) 
+
+      Let Z2Zp_canon_pos u : 0 <= u -> { v | 〘u〙=〘v〙 /\ 0 <= v < Z.of_nat p }.
+      Proof.
+        intros H1. 
+        rewrite Z2Zp_pos; auto.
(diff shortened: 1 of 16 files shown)


DmxLarchey:
Time limit for CI has been reached, ie more than 50 minutes of compilation time !!

yforster:
I'd be happy to merge this if it's finished. Or do we want to wait for the journal paper to be finished?

DmxLarchey:
You could merge ... it contains proofs of Luca's thm and Lagrange thm but maybe it is still lack the H10(nat) -> H10(Z) reduction

yforster:
I think I prefer merging now and then adding the `H10(Z)` reduction later (I'm on it already).

PR #27, opened 2020-01-14
Informative lift of the definition of reduces, i.e. ⪯

DmxLarchey:
This is a proposed PR for a change in the definition of `reduces` in `Problems/Reduction.v` to make it **informative**. Indeed, the original definition was:

```
 Definition reduces X Y (p : X -> Prop) (q : Y -> Prop) := 
    exists f : X -> Y, forall x, p x <-> q (f x).
```

and the proposed new definition is now:

```
Definition reduces X Y (p : X -> Prop) (q : Y -> Prop) :=
    { f : X -> Y | forall x, p x <-> q (f x) }.
```
so just the replacement of non-informative `∃` with informative `∑`.

I acknowledge that this is a **fundamental change** that implied some changes in the code elsewhere, mostly in the `L/*` branch of the library that relied too much on non-informative features, so I had to lift several definitions in `L/Computability/Synthetic.v` to informative as well, i.e. `L_decidable`, `L_enumerable`, `L_recognisable`, `L_enumerable_recognisable`, `L_enum`, `projection`, `L_enumerable_ext`, `L_enumerable_enum`, `L_enumerable_halt`. For each definition/lemma like e.g. `lem`, I did create an informative version named `lem_t` and proved the equivalence 

```
lem_t_eq ... : inhabited(lem_t ...) <-> lem ...
``` 

In the other developments like `ILL` or `H10` or `FOL` ... changes where minimal and proof scripts just worked out of the box. Notice however that some proof scripts do not explicitly name `intros` variables, and names picked up by Coq are not stable when migrating from `Prop` to `Type`. In general, I would advise to **never** use a name which was not introduced explicitly in the proof script.

Overall, the change worked out quite smoothly and this PR proposes a complete migration to informative reduction of the `master` branch. 

The change is motivated by three reasons:
- first informative reductions are needed/helpfull in the developments on decidability of finite FO satisfiability, including the full Trakhtenbrot theorem.
- second, it seems right that a reduction should actually be represented by a type with computational content, hence it can be extracted.
- the third reason is that the informative version allows to use parametrized reductions w/o having to use dependent choice. See below.

Indeed, when you have a _parametrized reduction_ like `∀ t : T, p t ⪯ q t` with `X Y : T -> Prop`, `p : ∀ t : T, X t -> Prop` and `q : ∀ t : T, Y t -> Prop`, one trivially can get a map 
```
{ f : ∀t, X t -> Y t | ∀ t x, p t x <-> q y (f t x) }
``` 
in the informative case but it is not possible to establish 
```
∃ (f : ∀t, X t -> Y t), ∀ t x, p t x <-> q y (f t x)
``` 
in the non-informative case without using dependent choice, more precisely [`DependentFunctionalChoice`](https://coq.inria.fr/library/Coq.Logic.ChoiceFacts.html).

Also the PR proposes some non-destructive add-ons in the `Shared/Libs/DLW/*` libraries.

### DIFF
--- a/theories/FOL/Reductions.v
+++ b/theories/FOL/Reductions.v
@@ -5,13 +5,15 @@ From Undecidability Require Export Problems.Reduction DecidableEnumerable.
 Lemma dec_red X (p : X -> Prop) Y (q : Y -> Prop) :
   p ⪯ q -> decidable q -> decidable p.
 Proof.
-  intros [f] [d]. exists (fun x => d (f x)). intros x. rewrite H. eapply H0.
+  intros [f Hf] [d Hd]. 
+  exists (fun x => d (f x)). 
+  intros x; rewrite <- Hd; apply Hf.
 Qed.
 
 Lemma red_comp X (p : X -> Prop) Y (q : Y -> Prop) :
   p ⪯ q -> (fun x => ~ p x) ⪯ (fun y => ~ q y).
 Proof.
-  intros [f]. exists f. intros x. now rewrite H.
+  intros [f Hf]. exists f. intros x. now rewrite Hf.
 Qed.
 
 Section enum_red.
@@ -61,10 +63,10 @@ Theorem not_decidable X Y (p : X -> Prop) (q : Y -> Prop) :
   p ⪯ q -> enumerable__T X -> ~ enumerable (compl p) ->
   ~ decidable q /\ ~ decidable (compl q).
 Proof.
-  intros. split; intros ?.
-  - eapply H1. eapply dec_red in H2; eauto.
-    eapply dec_compl in H2. eapply dec_count_enum; eauto.
-  - eapply H1. eapply dec_red in H2; eauto.
+  intros H1 H2 H3; split; intros H4.
+  - eapply H3. eapply dec_red in H4; eauto.
+    eapply dec_compl in H4. eapply dec_count_enum; eauto.
+  - eapply H3. eapply dec_red in H4; eauto.
     eapply dec_count_enum; eauto. now eapply red_comp.
 Qed.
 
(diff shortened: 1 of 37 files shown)


yforster:
I am unfortunately not in favour of this change. 

First, the question immediately extends to other notions like decidability and enumerability. Traditionally you can prove statements like `nontrivial q -> decidable p -> p ⪯ q` and with our current definitions we can in fact prove this. Once we move to an informative definition of `⪯` we would have to move the rest as well. And here I believe it's completely clear that notions like decidability and enumerability are something logical, and considering that a problem for us is a function `P -> Prop`, being something logical means to live in `Prop`.

Second, I believe that one can carry out synthetic computability theory in settings with classical assumptions like logical excluded middle (I'm aware that I first have to convince sceptics by a proof that this is okay ;) ). If we move to informative definitions, that's impossible in future, even if we can prove that excluded middle still preserves the computability of all definable functions. It's also possible that I'm proven wrong in my conjecture, then making reducibility informative is obviously the right way to go (and I owe several people a beer or two).

I think in general the situation is similar to the definition of decidability via dependently typed function or like we do currently. Reasoning based on the definition `forall x, {P x} + { ~ P x}` is nicer in many cases, and I well believe that reasoning using an informative definition of reducibility is nicer as well, but one can separate between reasoning in proofs (where we use the dependently typed version) and the definition exposed in theorems (where we use a propositional version).

So I'm very happy to include a notion of "informatively reducible" to the library, and whenever we can we prove this first and only as a second step deduce reducibility as currently defined. But the "real" definition has to be propositional I think.

DmxLarchey:
You mean decidable P is `exists f : X -> bool, forall x, P x <-> f x = true` which is equivalent to `inhabited (forall x, { P x } + { ~ P x })` ? It turns out I usually use the typed version of decidable ie either `{ f : X -> bool |   forall x, P x <-> f x = true }` or `forall x, { P x } + { ~ P x }`.

The main problem for me is that in the logical case, if you have a dependent decider `forall i : I, decidable (P i)` over `P : forall i, X i -> Prop`, then you cannot turn it into `exists (f : forall i, X i -> bool), forall i (x : X i), P i x <-> f i x = true` w/o the axiom of dependent choice whereas you can easily do it in the dependent case.

So for Trakhtenbrot for instance, we really use this tricks that works only in the informative case.

So we shall discuss option to have both worlds together and possibly downgrade the informative case to logical ...

DmxLarchey:
That is we have to review all the definitions that exists and write a comprehensive summary of their equivalences ... in one of the main files

DmxLarchey:
I have split the PR into on big DLW/libs update and a smaller informative update that needs to be converted to both informative and non-informative reductions. So far, all the reductions are informative, hence non-informative as well.

DmxLarchey:
Closing the PR to be resubmitted later as two PRs

PR #30, opened 2020-02-07
Undecidability of first-order ZF

dominik-kirst:
The contributed files consist of a mechanisation of the undecidability of semantic entailment in first-order ZF. The file Problems/ZF.v contains the axiomatisation and the statement of ZF-entailment and the file Reductions/BPCP_to_ZF.v contains a verified many-one reduction from the Post correspondence problem.


DmxLarchey:
Hi Dominik,

It is my understanding that the signature of ZF is (ø,{∈²}) possibly with an identity

https://en.wikipedia.org/wiki/Zermelo%E2%80%93Fraenkel_set_theory

You seem to have emptyset, pairs etc in your signature ... what would it entail not having those ?

dominik-kirst:
Hi Dominique,

both versions of ZF are equivalent (e.g. interreducible), in general ZF plus definable extensions is conservative over pure ZF. Practically, the encoding of PCP and the verification would be a more involved (but of course doable) if the set operations weren’t present as functions.

Best wishes,
Dominik

DmxLarchey:
> Hi Dominique, both versions of ZF are equivalent (e.g. interreducible), in general ZF plus definable extensions is conservative over pure ZF. 

Is this true **constructivelly** ? Remember that it is not easy to recover functions from relations with arbitrary types ...

dominik-kirst:
Yes, with the right definitions it will be constructive ;-)

For instance we could require that ZF model satisfy unique choice or at least contain the set operations as functions.

DmxLarchey:
Hi Dominik,
 
Could you did it with a FO axiomatization instead of semantic models ?

Best,

D.

dominik-kirst:
Yes, this should work as well. However, given a ZF deduction you would probably anyways argue semantically to reconstruct a PCP solution and also in the converse direction I guess one would rather use completeness to get a ZF deduction instead of constructing the derivation syntactically.

Dominik

dominik-kirst:
In fact, implementing the undecidability for the ZF deduction system (mentioned by @DmxLarchey ) was not too much of a pain, see Reductions/BPCP_ZFD.v. The preservation direction (BPCP B -> ZF proves (solvable B)) could in principle be established using completeness, but this would require either classical axioms or a very careful (i.e. reifiable) semantic proof. I instead directly translated the semantic proof into the deduction system.

The reflection direction (ZF proves (solvable B) -> BPCP B) is easily argued via a suitable soundness theorem and reuses the semantic reflection. However, this allows only an intuitionistic deduction system since classical rules cannot be shown sound for the standard model constructively. I guess instead of using soundness one could as well argue via cut-elimination and proof analysis but this would probably be more complicated.

DmxLarchey:
Hi Dominik,

> In fact, implementing the undecidability for the ZF deduction system (mentioned by @DmxLarchey ) was not too much of a pain, see Reductions/BPCP_ZFD.v.... I instead directly translated the semantic proof into the deduction system.

I somehow expected that you would encode this part in the deduction system directly. However, the deduction system you propose in not exactly that of ZF since you have syntax for pair, emptyset, union and powerset. How would it go w/o terms ? More generally, could you be able to show the equivalence between the rich signature (with terms) and the minimal signature (with only membership) in a constructive meta-theory that does not assume XM and choice ? Could you do it w/o going through the semantics, ie by proof transformations ? Of course, one direction is more obvious than the other.

> The reflection direction (ZF proves (solvable B) -> BPCP B) is easily argued via a suitable soundness theorem and reuses the semantic reflection. However, this allows only an intuitionistic deduction system since classical rules cannot be shown sound for the standard model constructively. I guess instead of using soundness one could as well argue via cut-elimination and proof analysis but this would probably be more complicated.

Yep and this reflects with the paper about a constructive approach to FOL. Validity and provability are two different things even if they are both undecidable ... Cut-elemination for 1st order ZF ? would this hold ? Right now your proof system for ZF is Hilbert style, is it not ? And this proof system includes XM which is not sound constructivelly. Could you make it sound by requiring the model to satisfy XM w/o assuming XM in the meta-theory ?

Best, 

Dominique

dominik-kirst:
Yes, I think that the proof transformations for the equivalence of both representations of ZF should go through constructively. This should extend to any definable extension of the syntax.

The proof system I use is a standard natural deduction system in finite contexts with the natural extension to infinite contexts. I use the intuitionistic version as it is strong enough to trace the encoded PCP solution and avoids the soundness problem in the other direction. As you say, a classical deduction system is sound when restricting to only classical models, however, this does not help since the standard model used in the proof is not classical.

PR #33, opened 2020-02-22
Fix makefile and coqdoc compilation

lgaeher:
This restores the makefile before 6ddce394ca5bc2206f6e0e2bcecc0582e70fb6a3 (auto-generated makefiles probably should not be in the repo?) and fixes compilation of coqdocs. 

The removal of the parse-comments flag is debatable, but with it, the docs don't build (lexer errors). I'm not sure about that, it might be better to fix the underlying problem, but coqdoc isn't very generous with debug information, even when using the verbose flag.


DmxLarchey:
I definitely agree that having an auto-generated file in the repo, all the more in the `master` branch, is not a good thing. The new `theories/Makefile` looks fine for me since it allow to compile eg `make H10/DPRM.vo` from the `theories` directory.

Regarding the `coqdoc` part, I am not the best to judge. Yannick, do you have an opinion ?

PR #39, opened 2020-03-31
PCP refactoring into PCP, SR, GFG

mrhaandi:
- refactored PCP into PCP, SR, CFG according to discussed design guidelines
- removed Shared.Prelim from Problems.Reduction


yforster:
Thanks, that's a great first step towards a better shape of the library. I'm happy to merge. @DmxLarchey any objections?

DmxLarchey:
Ok for 8.11 but when shall we Zoom/Skype? We have to be on sync on where we are going ...

PR #40, opened 2020-04-10
Less Warnings, Improved Compilation Speed

mrhaandi:
- Eliminated ~500 warnings in `PCP`, `CFG`, `SR`, `MM`, `H10C`, `H10`, `MuRec`, `ILL`, and `HOU`
- Improved overall compilation speed by about 10%
- Improved compilation speed of `PCP`, `CFG`, and `SR` by one order of magnitude

PR #41, opened 2020-04-11
Proposal to integrate the Trakhtenbrot code (IJCAR 2020)

DmxLarchey:
The code contains:

- update of the `Shared/Libs/DLW` library;
- informative `⪯ᵢ` and non-informative `⪯` reductions;
- the `TRAKHTENBROT` code with informative reductions.

For the moment, no description in `Problems` and `Reductions`. The file `TRAKHTENBROT/summary.v` lists the main results of the [IJCAR paper](https://members.loria.fr/DLarchey/files/papers/trakhtenbrot.pdf).


DmxLarchey:
@yforster : I am not able to add @mrhaandi neither as reviewer nor a assignee

DmxLarchey:
Ok I added @mrhaandi in the invited people (as a reader only).

mrhaandi:
1. Is there a reason why an older, redundant `PCP.v` was added to `/Problems`? This contradicts the decision to remove the `/Problems` in the long run. The current binary PCP characterization by derivability is `dPCPb` in `/PCP/PCP.v`.
2. Why `ireduces` instead of `informatively_reduces`? A descriptive name suits better a key definition that is used in main theorems.
3. At some point (here seems appropriate) the notions `enumerable`, `decidable`, `undecidable` should be declared at the same level as `reduces`.

DmxLarchey:
>     1. Is there a reason why an older, redundant `PCP.v` was added to `/Problems`? This contradicts the decision to remove the `/Problems` in the long run. The current binary PCP characterization by derivability is `dPCPb` in `/PCP/PCP.v`.

Well there are results in there that did not seem to occur in the `PCP` lib such as `derivable_BPCP` and `BPCP_derivable` (I know the names should be `derivable_PCP` and `BPCP_derivable`). Anyway, where should the results go in the new structure ... I think we need a lib. showing that all these characterisations are equivalent. Also, I do not see why

```
Definition dPCPb {X} : stack X -> Prop :=
  fun P => exists u, @derivable X P u u.
```

does not occur ?

>     2. Why `ireduces` instead of `informatively_reduces`? A descriptive name suits better a key definition that is used in main theorems.

Ok I can change this very quickly 

>     3. At some point (here seems appropriate) the notions `enumerable`, `decidable`, `undecidable` should be declared at the same level as `reduces`.

Here there are several notions for enumerable. We should discuss them. Also, I am not in favour `undecidable` because any definition we give in the synthetic setting is too restrictive. We could call it `syn_undec` perhaps ?

DmxLarchey:
For `enumerable` for types we have

```
Definition type_enum := exists f : nat -> option X, forall x, exists n, Some x = f n.

Definition list_enum := exists f : nat -> list X, forall x, exists n, In x (f n).
```

but these two are equivalent. Of course there are informative variants.

But for `enumerable` w.r.t predicates, we have 

```
Definition opt_enum P := exists f : nat -> option X, 
                           forall x, P x <-> exists n, Some x = f n.

Definition rec_enum P := exists (Q : nat -> X -> bool),
                           forall x, P x <-> exists n, Q n x = true.
```
which are not generally equivalent. Of course there are also two informative versions ...

`opt_enum` and `rec_enum` are not equivalent unless the type `X` is both `discrete X` and `type_enum X` hold, ie. `X` is a _datatype_.

mrhaandi:
@DmxLarchey 
> Well there are results in there that did not seem to occur in the `PCP` lib such as `derivable_BPCP` and `BPCP_derivable`. I think we need a lib. showing that all these characterisations are equivalent. 

During refactoring I removed no facts about PCP and collected facts about different PCP variants in the PCP folder (either in `Util` if it is a general property, or in `Reductions` if it involves multiple PCP variants). E.g. you can find `derivable_BPCP` in `./PCP/Reductions/PCPb_iff_dPCPb.v`.
Also there it is shown that the characterizations are equivalent in 
```Lemma PCPb_iff_dPCPb P : PCPb P <-> dPCPb P.```

> About `Definition dPCP {X} : stack X -> Prop := fun P => exists u, @derivable X P u u.`

As seed `X := bool` subsumes any other undecidable variant. Is the case with a general `X` (previously not used) relevant?

**Comment by DmxLarchey** [comment]
@mrhaandi : I did remove `Problems/PCP.v` and dispatched it in `PCP/*`

> As seed `X := bool` subsumes any other undecidable variant. Is the case with a general `X` (previously not used) relevant?

Well most of the definitions hold for an arbitrary type `X` so I see no point in not factoring it out instead of having duplicate definitions for `nat` and `bool`.  So I defined `PCPX {X} P := exists u, @derivable X P u u` and `PCP := @PCPX nat` and `PCPb := @PCPX bool`.

**Comment by mrhaandi** [comment]
> So I defined `PCPX {X} P := exists u, @derivable X P u u` and `PCP := @PCPX nat` and `PCPb := @PCPX bool`.

Ok. Now I'm mostly fine with the PR. A minor remark:
```
Inductive BPCP (P : stack bool) : Prop := 
  | cBPCP : forall u, derivable P u u -> BPCP P.
```
This is highly redundant with `dPCPb` as `cBPCP` is a trivial Lemma about `dPCPb`.
Do you actively use `cBPCP` as constructor of an inductive type?

DmxLarchey:
> ```
> Inductive BPCP (P : stack bool) : Prop := 
>   | cBPCP : forall u, derivable P u u -> BPCP P.
> ```
> 
> This is highly redundant with `dPCPb` as `cBPCP` is a trivial Lemma about `dPCPb`.
> Do you actively use `cBPCP` as constructor of an inductive type?

Yes indeed `cBPCP` is just a specialization of the `exist` constructor of `ex` but I wanted `BPCP` to be defined formally using an inductive rule, as is done in the paper.

yforster:
> But for `enumerable` w.r.t predicates, we have
> 
> ```
> Definition opt_enum P := exists f : nat -> option X, 
>                            forall x, P x <-> exists n, Some x = f n.

In my opinion, this is the "right" definition of enumerability.

> 
> Definition rec_enum P := exists (Q : nat -> X -> bool),
>                            forall x, P x <-> exists n, Q n x = true.
> ```

I would not call this enumerability, but "recognisability" or (better in my opinion) semi-decidability. I would switch the order on `Q` here: Traditionally, a semi-decider is a partial function on `X` returning something computationally irrelevant, so we can say it's a partial function `X -> unit`. So `Q` should be `X -> nat -> bool`, since `nat -> bool` in a way is a partial unit value.

By renaming it's more clear that the notions are not equivalent in general.

Very happy with all the rest!

We could use the definition of BPCP by Dominique as the definition of `dPCPb`? It's probably not even breaking proof scripts.

mrhaandi:
I agree with Yannick on the topic of `enumerable`.
>We could use the definition of BPCP by Dominique as the definition of dPCPb? It's probably not even breaking proof scripts.

Yes, this also would be fine.

yforster:
I'm in favour of merging this before sorting out the right place to define `enumerable` and the like. That way we can merge #40 quickly afterwards and I can have a look at the warnings in `L` which do not correspond to automation.

DmxLarchey:
@yforster:

> I would not call this enumerability, but "recognisability" or (better in my opinion) semi-decidability.

We already had such discussion on the `H10` paper and in principle I agree with you. Indeed, `opt_enum` contains this idea of enumeration by `nat`. Naively, one would even pick the direct image of some `f : nat -> X` but this would not work for empty predicates...

But let us not forget that there is established and widely used terminology and for instance _recursively enumerable for `P : nat -> Prop`_ usually means that `P` is the termination predicate of a partial recursive function, equivalently by Kleene's normal form theorem, there is a primitive recursive function `f : nat -> nat -> bool` such that `P n <-> exists w, f n w = true` which corresponds nicely to `rec_enum`. Of course `nat` is a data-type which means, this is constructively equivalent to `opt_enum`.

So I suggest we think about it for a while so as to fix a terminology which is somehow compatible with existing terminology. Hence, let us collect all ideas so that we can make well thought of decision.

I also like _semi-decidable_ which stresses that there is a way to know for sure when a element belongs to `P` while it might not be possible to know when elements do not belong to `P`.

yforster:
I moved the discussion to #42 because the PR now disappeared from the list of open PRs.

PR #43, opened 2020-04-14
Remove more warnings in L and TM part

yforster:
4 kinds of warnings left:

- `ambiguous-paths`, stemming from coercions declared in the base library. I have already pushed a fix to the base library, but it's not in the opam package of the base library yet.
- `implicit-core-hint-db`: Only in `Shared/DLW/Libs`. I didn't want to interfere with the open PR extending there, so I'll leave those warnings for @DmxLarchey.
- `notation-overridden`: We are re-defining lots of notations in the `FOL` part of the project. After that's unified I hope that the problem just disappears. For now disabled via `_CoqProject`
- `unused-intro-pattern`: This warning is generated in the `TMcorrect` tactic, but I don't know where. Maybe @mwuttke97 has an idea how to fix this. For now disabled via `_CoqProject`.

### DIFF
--- a/CeCILL_LICENSE.txt
+++ b/CeCILL_LICENSE.txt
@@ -1,5 +1,7 @@
 
-CeCILL FREE SOFTWARE LICENSE AGREEMENT
+  CeCILL FREE SOFTWARE LICENSE AGREEMENT
+
+Version 2.1 dated 2013-06-21
 
 
     Notice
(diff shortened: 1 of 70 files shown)


mrhaandi:
Good work. Although not a warning, what about the very long (and to me incomprehensible) output around lines 775-1412 in the worker build log?

yforster:
There are two issues which I'd still like to fix:

- one is lots of `Compute` and `Check` statements in various places, mainly 
in `TM`. They should just be removed.
- another one is that the `TMcorrect` tactic has lots of debut output. I 
disabled parts of it, but since the tactic is defined in various places 
using smpl that's not completely straightforward. I'll consult with 
Maximilian what's the best way to go there

mwuttke97:
The "unused intropattern" probably comes from the tactic `destruct_shelve` in `Combinators/Combinators.v`. It's a very uggly tactic, indeed.
The purpose of the tactic is to make a 'complete' case analysis over a (finite) label, `shelv`ing when further destructing is not possible any more. This is used in order to create the relation the shape for the relation of a `Switch` machine. (See `Ltac smpl_match_Realise`.)

DmxLarchey:
@yforster could we update the CeCill License ?

There is also a problem with global `Makefile` that does not transmit `-j 4` option (parallel compilation) to its sub-processes. To get parallel compile, I need to move in `theories` and type

```
make -j 4 -f Makefile.coq all
```

in there...

yforster:
The license is already updated to 2.1 in this branch, I copy-pasted frome `master`.

Regarding `make` I can't reproduce. Typing `make -j 4` in the top-level directory starts parallel compilation, as indicated by the `+make` in the `Makefile` (the `+` means "propagate options"). Maybe it's different versions of `make`? I have

```
GNU Make 4.3
Built for x86_64-pc-linux-gnu
```

yforster:
Thanks Maximilian! I removed more debug printing in `destruct_shelve` and I think I understand the "unused-intro-pattern" warning now. It can be reproduced by compiling 

```
Lemma test : forall t : unit, True.
Proof.
  intros t.
  destruct t as [ x ].
  tauto.
Qed.
```
which produces `Warning: Unused introduction pattern: x [unused-intro-pattern,tactics]`.

I don't see a way to rewrite the tactic such that it doesn't produce the warning. But it seems harmless to disable the warning, so we'll just keep it disabled.

If CI agrees, I'd merge this PR then.

DmxLarchey:
> ```
> GNU Make 4.3
> Built for x86_64-pc-linux-gnu
> ```

Mine is `GNU make 4.1` ... these are the forwarded processes:

```
dominique@napo:~/GitHub/coq-library-undecidability$ ps aux | grep make | cat
dominiq+  6217  0.0  0.0   9688  2268 pts/18   S+   10:49   0:00 make -j 4 all
dominiq+  6218  0.0  0.0   9764  2440 pts/18   S+   10:49   0:00 make -C theories all
dominiq+  6219  0.0  0.0  15516  8360 pts/18   S+   10:49   0:00 make -f Makefile.coq all
dominiq+  6236  0.0  0.0   4496   704 pts/18   S+   10:49   0:00 /bin/sh -c make --no-print-directory -f "Makefile.coq" real-all
dominiq+  6237  0.0  0.0  15952  8792 pts/18   S+   10:49   0:00 make --no-print-directory -f Makefile.coq real-all
dominiq+  7153  0.0  0.0  14260   968 pts/19   S+   10:59   0:00 grep --color=auto make
```
and `-j 4` is not transmitted on the command line. Could it be transmitted through environment variables? Because indeed there are 4 `coqc` processes running in parallel.

yforster:
Strange, maybe it is done via environment variables. I don't know much about `make` (next to nothing actually), I just googled at some point how to propagate and the answer on stackoverflow or something was "use `+make`".

I'm merging this PR now to make it easier for others to work in parallel. We can discuss in the issue w.r.t. Checks and then maybe remove some of them.

PR #45, opened 2020-04-17
Unified definition of first-order logic syntax

yforster:
Here's the promised suggestion which unified syntax of FOL we might use in the future. As discussed in the Zoom meeting I tried to incorporate all the improvements in the different projects.

My idea would be that as a first step we try to base all 3 existing developments on this file (plus a couple of shared files defining the instances for frequent syntax combinations, e.g. FOL with all quantifiers and all binary symbols). For this first step we can keep two lemma bases for substitution: One will be essentially the one of Autosubst to have `asimpl` working (and will likely depend on Fext), another one is the one from the Trakhtenbrot project. Similarly, we can start by having different deduction systems and only unify them in a second step.

I created this as draft PR where we can discuss. It probably makes sense that we first all agree on a new definition of Syntax, port out projects and then go to step 2 and unify other parts (but that plan is also up for discussion).

### DIFF
--- a/theories/FOLP/Input.v
+++ b/theories/FOLP/Input.v
@@ -177,13 +177,13 @@ Module minimal.
   Coercion V := @bVar min.
 
   Definition phi1 :==: ∀ x, !P x x.
-  Print phi1.
+  (* Print phi1. *)
 
   Definition phi2 :==: free y z, ∀ x, !P x y --> !P z x.
-  Print phi2.
+  (* Print phi2. *)
       
   Definition phi3 :==: ∀ x y, !P ($f1 $e) y --> (∀ z, !P x z) --> !Q.
-  Print phi3.
+  (* Print phi3. *)
 End minimal.
   
 Notation "£ f" := (bEmbed f List.nil) (at level 2). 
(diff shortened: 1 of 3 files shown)


dominik-kirst:
I'm good with this, just two minor comments:

(1) The operators and quantors could be bundled to generate fever assumptions in a section

(2) I think there is a mistake in the definition of up, it should be
      Definition up (σ : nat -> term) := scons (var 0) (funcomp (subst_term (funcomp var S)) σ).

DmxLarchey:
Would it be a big issue to substitute `Vector.t` with `vec` from `Shared/Libs/DLW/Vec` and also use `pos` for finite indices ? Otherwise this change might largely impact the Trakhtenbrot devel... I could try to change `vec` so that the base type matches `Vector.t` syntactically (same for `pos` with `Fin.t`) but this might have a huge impact on many part of the library (`ILL`, `BSM`, `MM`, `H10` ...)

DmxLarchey:
Ok I did write PR #47 to unify `Vector.t` and `vec` by updating the `DLW` lib yo make it compatible with the standard library of Coq. The impact was not so strong but still some problems emerged. They are commented there.

DmxLarchey:
PR #47 has being merged into branch `coq-8.11`

yforster:
> I'm good with this, just two minor comments:
> 
> (1) The operators and quantors could be bundled to generate fever assumptions in a section
> 
> (2) I think there is a mistake in the definition of up, it should be
> Definition up (σ : nat -> term) := scons (var 0) (funcomp (subst_term (funcomp var S)) σ).

Thanks for the correction!

I bundled the operators into one typeclass, it might be easier to treat indeed.

If we are all happy with this I would next write down shared definitions of standard Tarski semantics and of provability. This also requires operator instances for the full definition and the `forall, impl`-fragment.

DmxLarchey:
Just a check. Does the inductive type of term depend on a signature of just functions or else on a whole FO signature, i.e. including relation symbols ? Because for Trakhtenbrot, I need the former case.

dominik-kirst:
> Just a check. Does the inductive type of term depend on a signature of just functions or else on a whole FO signature, i.e. including relation symbols ? Because for Trakhtenbrot, I need the former case.

Jep, the type term just depends on an instance of funcs_signature :)

yforster:
Superseded by #95.

PR #46, opened 2020-04-18
New central Synthetic.Definitions file

yforster:
I started reworking the files in the problem areas `Shared/Prelim.v`, `Problems/Reduction.v`, `FOL/DecidableEnumerable.v`, and `FOL/Reductions.v`.

What I propose is one central directory `Synthetic`, containing

- a file `Definitions.v` with definitions for all synthetic notions of decidability, enumerability, reduction and their informative counterparts.
- files `<Concept>Facts.v` for each of these concepts with the relevant facts previously somewhere in the `FOL` part

I also started splitting up `Shared/Prelim.v` and only import (no exports!) the relevant chunks in the Facts files. The only exported notation is the notation for reduction and informative reduction.

This PR won't compile, because I haven't adapted everything yet, but the `Synthetic` directory compiles as is.

I switched from `informatively_` as prefix to `inf_` because the former leads to annoyingly long names which are hard to type, but that's up for discussion.

### DIFF
--- a/theories/FOL/DecidableEnumerable.v
+++ b/theories/FOL/DecidableEnumerable.v
@@ -1,477 +1,477 @@
 (** * Decidability and Enumerability *)
 
 From Undecidability Require Export Shared.Prelim.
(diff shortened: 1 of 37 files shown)


yforster:
For the record: If travis should ignore a commit, one can put `[skip ci]` in the commit message.

PR #47, opened 2020-04-19
Unify Fin.t / pos and Vector.t / vec

DmxLarchey:
This is a PR to unify 

- `Fin.t` (from std lib) with `DLW/Vec/pos.pos`;
- `Vector.t` (from std lib) with `DLW/Vec/vec.vec`.

Not so many changes. For `pos` things went smoothly. For `vec`, it was more complicated because of several reasons:

- `Vector.cons X x n v` = `vec_cons X n x v` ie. arguments are not in the same order, which impacts the `induction` or `destruct` tactics with explicit intro patterns;
- `vec_cons` can no longer be seen as a constructor but the `_ ## _` notation can largely mitigate this;
- proof scripts that use auto-generated names are messed up because `induction v` will now behave as `induction v as [ | x n t ]` and not as `induction v as [ | n x v ]`. 

Might I advice everyone to _always introduce the names they use_ because correcting such migration bugs can become very complicated if the proofs are intricate.

### DIFF
--- a/theories/H10/Fractran/mm_fractran.v
+++ b/theories/H10/Fractran/mm_fractran.v
@@ -146,7 +146,7 @@ Proof with eauto; try omega.
         eapply primestream_divides in H0...
         eapply divides_encode_state in H0... 
       * eapply primestream_divides in H0... subst.
-        eapply (H n p). eauto.
+        eapply (H n t). eauto.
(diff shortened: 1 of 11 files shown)


yforster:
That's amazing! Glad that it wasn't a full day project and sorry for all the implicitly introduces names :)

PR #48, opened 2020-04-20
Refactored changes from private complexity-development 

fakusb:
This commit brings the TM and L-extraction/verification frameworks up-to-date with the features needed in my complexity theory development.

This is a Draft pull request as I want to first check the travis-output. Furthermore, I will look at removing more warnings and stripping unneeded files before I mark this as non-draft.


yforster:
I fixed the last warnings. If the CI run is also warning-free we can merge.

fakusb:
We should be able to merge now.

yforster:
I am merging this PR and will then rebase #46 on top of it.

However everybody please be aware that this PR relies on an updated MetaCoq version, available via our opam repo only (for the moment, the fix will be in the next official MetaCoq opam release).

For everybody working on the library, after `git pull` a `opam install . --deps-only` should fix the local setup and update the MetaCoq version as needed. Tagging the active contributors: @DmxLarchey @mrhaandi @dominik-kirst

DmxLarchey:
May I suggest `opam update` before `opam install . --deps-only` ...

But having branches depend on different versions of external libraries is going to make development in several branches at the same time potentially a real mess ... and the same for forks. On needs to have several `opam` switches I suppose ...

fakusb:
Yes, this is not optimal. We can switch back to the upstream-metacoq as soon as the merged PR https://github.com/MetaCoq/metacoq/pull/407 is released.

fakusb:
Although all other branches on 8.11 can just merge with the 8.11-branch and the opam-switch for 8.11 would be in sync again, as Yannick just mentioned to me.

DmxLarchey:
True but for "mirrored" branches on forks but shall we expect one `opam switch` for every active branch in the repo? I can understand the need to synchronize our developments but I do not think this should be done on the `default` branch which is now `coq-8.11`. The `default` branch should not depend on _alpha releases_ of external libraries.

May be creating a branch on top of `coq-8.11` on the main repo. would allow everyone to sync on that instead of the `default` distributed branch?

yforster:
There's not even a beta release of MetaCoq (and the "alpha" is rather arbitrary, it might as well have been `coq-metacoq.2.1`).

Running `opam update && opam install . --deps-only` on the `coq-8.11` branch will update the development environment s.t. it works for every other branch and PR on the same Coq version. In other words, updating the development environment should always be backwards compatible w.r.t. other branches. So for contributors that don't want to get into the business of managing external dependencies they never have to: As long as their environment is up to date with the requirements specified on the `coq-8.11` branch, they are always good to work on any other branch.

PR #50, opened 2020-04-21
New central Synthetic.Definitions file

yforster:
(I force pushed on the branch and github got confused, so reopening this PR)

I started reworking the files in the problem areas `Shared/Prelim.v,` `Problems/Reduction.v`, `FOL/DecidableEnumerable.v`, and `FOL/Reductions.v`.

What I propose is one central directory Synthetic, containing

- a file `Definitions.v` with definitions for all synthetic notions of decidability, enumerability, reduction and their informative counterparts.
- files `<Concept>Facts.v` for each of these concepts with the relevant facts previously somewhere in the FOL part
- a file `Undecidability.v` containing the definition of undecidability. This is not in the `Definitions` file to avoid a compile-dependency on Turing machines (and thus the base lib) for the early parts of the library.

I also started splitting up Shared/Prelim.v and only import (no exports!) the relevant chunks in the Facts files. The only exported notation is the notation for reduction and informative reduction.

This PR won't compile, because I haven't adapted everything yet, but the Synthetic directory compiles as is.

I switched from `informatively_` as prefix to `inf_` because the former leads to annoyingly long names which are hard to type, but that's up for discussion.

### DIFF
--- a/theories/CFG/CFP_undec.v
+++ b/theories/CFG/CFP_undec.v
@@ -3,23 +3,22 @@ Require Import Undecidability.CFG.CFP.
 From Undecidability.CFG.Reductions Require 
   PCP_to_CFPP PCP_to_CFPI.
 
-Require Import Undecidability.Problems.TM.
-Require Import Undecidability.Problems.Reduction.
+Require Import Undecidability.Synthetic.Undecidability.
 
 Require Undecidability.PCP.PCP_undec.
 
 (** The Context-free Post Grammar Palindrome Problem is undecidable. *)
-Lemma CFPP_undec : HaltTM 1 ⪯ CFPP.
+Lemma CFPP_undec : undecidable CFPP.
 Proof.
-  eapply reduces_transitive.
+  eapply undecidability_from_reducibility.
   exact PCP_undec.PCP_undec.
   exact PCP_to_CFPP.reduction.
 Qed.
 
 (** The Context-free Post Grammar Intersection Problem is undecidable. *)
-Lemma CFPI_undec : HaltTM 1 ⪯ CFPI.
+Lemma CFPI_undec : undecidable CFPI.
 Proof.
-  eapply reduces_transitive.
+  eapply undecidability_from_reducibility.
   exact PCP_undec.PCP_undec.
   exact PCP_to_CFPI.reduction.
 Qed.
--- a/theories/CFG/CFG_undec.v
+++ b/theories/CFG/CFG_undec.v
@@ -3,15 +3,14 @@ Require Import Undecidability.CFG.CFG.
 From Undecidability.CFG.Reductions Require 
   CFPP_to_CFP CFPI_to_CFI.
 
-Require Import Undecidability.Problems.TM.
-Require Import Undecidability.Problems.Reduction.
+Require Import Undecidability.Synthetic.Undecidability.
 
 Require Undecidability.CFG.CFP_undec.
 
 (** The Context-free Palindrome Problem is undecidable. *)
-Lemma CFP_undec : HaltTM 1 ⪯ CFP.
+Lemma CFP_undec : undecidable CFP.
 Proof.
-  eapply reduces_transitive.
+  eapply undecidability_from_reducibility.
   exact CFP_undec.CFPP_undec.
   exact CFPP_to_CFP.reduction.
 Qed.
@@ -19,9 +18,9 @@ Qed.
 Check CFP_undec.
 
 (** The Context-free Intersection Problem is undecidable. *)
-Lemma CFI_undec : HaltTM 1 ⪯ CFI.
+Lemma CFI_undec : undecidable CFI.
 Proof.
-  eapply reduces_transitive.
+  eapply undecidability_from_reducibility.
   exact CFP_undec.CFPI_undec.
   exact CFPI_to_CFI.reduction.
 Qed.
--- a/theories/FOL/DecidableEnumerable.v
+++ b/theories/FOL/DecidableEnumerable.v
@@ -1,477 +1,477 @@
 (** * Decidability and Enumerability *)
 
 From Undecidability Require Export Shared.Prelim.
+From Undecidability.Synthetic Require Export Definitions DecidabilityFacts EnumerabilityFacts ListEnumerabilityFacts MoreEnumerabilityFacts.
 
+(* (* ** Definitions *) *)
 
-(* ** Definitions *)
+(* Definition compl X p := fun x : X => ~ p x. *)
+(* Definition decidable {X} (p : X -> Prop) := exists f, forall x, p x <-> f x = true. *)
+(* Definition enumerable {X} (p : X -> Prop) := exists f, forall x, p x <-> exists n : nat, f n = Some x. *)
 
-Definition compl X p := fun x : X => ~ p x.
-Definition decidable {X} (p : X -> Prop) := exists f, forall x, p x <-> f x = true.
-Definition enumerable {X} (p : X -> Prop) := exists f, forall x, p x <-> exists n : nat, f n = Some x.
+(* Definition discrete X := decidable (fun '(x,y) => x = y :> X).  *)
+(* Definition enumerable__T X := exists f : nat -> option X, forall x, exists n, f n = Some x. *)
 
-Definition discrete X := decidable (fun '(x,y) => x = y :> X). 
-Definition enumerable__T X := exists f : nat -> option X, forall x, exists n, f n = Some x.
+(* (* *** more practical type-theoretic characterisations *) *)
 
-(* *** more practical type-theoretic characterisations *)
+(* Lemma dec_decidable' X p : *)
+(*   (forall x : X, dec (p x)) -> { f : _ | forall x, p x <-> f x = true}. *)
+(* Proof. *)
+(*   intros d. exists (fun x => if d x then true else false). intros x. destruct (d x); firstorder congruence. *)
+(* Qed. *)
 
-Lemma dec_decidable' X p :
-  (forall x : X, dec (p x)) -> { f : _ | forall x, p x <-> f x = true}.
-Proof.
-  intros d. exists (fun x => if d x then true else false). intros x. destruct (d x); firstorder congruence.
-Qed.
+(* Lemma decidable_iff X p : *)
+(*   decidable p <-> inhabited (forall x : X, dec (p x)). *)
+(* Proof. *)
+(*   split. *)
+(*   - intros [f]. econstructor. intros x. specialize (H x). destruct (f x); firstorder congruence. *)
+(*   - intros [d]. eapply dec_decidable' in d as [f]. now exists f. *)
+(* Qed. *)
 
-Lemma decidable_iff X p :
-  decidable p <-> inhabited (forall x : X, dec (p x)).
-Proof.
-  split.
-  - intros [f]. econstructor. intros x. specialize (H x). destruct (f x); firstorder congruence.
-  - intros [d]. eapply dec_decidable' in d as [f]. now exists f.
-Qed.
+(* Lemma discrete_iff X : *)
+(*   discrete X <-> inhabited (eq_dec X). *)
+(* Proof. *)
+(*   split. *)
+(*   - intros [] % decidable_iff. econstructor. intros x y; destruct (X0 (x,y)); firstorder. *)
+(*   - intros [d]. eapply decidable_iff. econstructor. intros (x,y). eapply d. *)
+(* Qed. *)
 
-Lemma discrete_iff X :
-  discrete X <-> inhabited (eq_dec X).
-Proof.
-  split.
-  - intros [] % decidable_iff. econstructor. intros x y; destruct (X0 (x,y)); firstorder.
-  - intros [d]. eapply decidable_iff. econstructor. intros (x,y). eapply d.
-Qed.
+(* (* ** Facts *) *)
 
-(* ** Facts *)
+(* Lemma dec_compl X p : *)
+(*   decidable p -> decidable (fun x : X => ~ p x). *)
+(* Proof. *)
+(*   intros [f]. exists (fun x => negb (f x)). *)
+(*   intros. rewrite H. destruct (f x); split; cbn; congruence. *)
+(* Qed. *)
 
-Lemma dec_compl X p :
-  decidable p -> decidable (fun x : X => ~ p x).
-Proof.
-  intros [f]. exists (fun x => negb (f x)).
-  intros. rewrite H. destruct (f x); split; cbn; congruence.
-Qed.
+(* Lemma dec_conj X p q : *)
+(*   decidable p -> decidable q -> decidable (fun x : X => p x /\ q x). *)
+(* Proof. *)
+(*   intros [f] [g]. exists (fun x => f x && g x). *)
+(*   intros. rewrite H, H0. destruct (f x), (g x); cbn; firstorder congruence. *)
+(* Qed. *)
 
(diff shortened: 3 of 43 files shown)


yforster:
@DmxLarchey @mrhaandi What do you think? I'd be happy to merge sooner rather than later

DmxLarchey:
My main concern is that we make it clear that we define _synthetic_ version which are supposed to be stronger that the regular notion as e.g. `L` defined. These notions do not allow to prove meta-theorem about decidability or undecidability in general because they do not capture them entirely. It would be good if the names actually reflect that aspect.

yforster:
> My main concern is that we make it clear that we define synthetic versions which are supposed to be stronger that the regular notion as e.g. L defined. These notions do not allow to prove meta-theorem about decidability or undecidability in general because they do not capture them entirely. It would be good if the names actually reflect that aspect.

Mh, I don't think I agree, but maybe that's not so important for now.

The PR doesn't actually introduce new names, it just brings all the definitions to a central place keeping the names the same (more or less).

We can maybe discuss naming issues in the next developers' Zoom call if there is need, but I'd be in favour of merging this re-ordering of files independently from the concrete terminology.

theories/CFG/CFP_undec.v
```diff
@@ -3,23 +3,22 @@ Require Import Undecidability.CFG.CFP.
...
 
-Require Import Undecidability.Problems.TM.
-Require Import Undecidability.Problems.Reduction.
+Require Import Undecidability.Synthetic.Undecidability.
 
 Require Undecidability.PCP.PCP_undec.
 
 (** The Context-free Post Grammar Palindrome Problem is undecidable. *)
-Lemma CFPP_undec : HaltTM 1 ⪯ CFPP.
+Lemma CFPP_undec : undecidable CFPP.
```
DmxLarchey, line 11: Here for instance you replace a statement about a reduction with a statement about `undecidable`. They are they same formally because we accept `HaltTM 1 ⪯ P`as a definition of (synthetic) undecidability.

DmxLarchey:
> Mh, I don't think I agree

Not sure on what you disagree but last Zoom time when we discussed informative reductions, you said that while you agree they can be used to show undecidability results by reduction (because they are stronger than non-informative reduction), using them to characterize undecidability leads to statements contradicting regular computability theory. Fair enough but what makes you think this would not also happen with non-informative synthetic undecidability ?

theories/CFG/CFP_undec.v
yforster: We would accept that, yes. But in the Zoom call we agreed on "We use `PCP_undec : decidable PCP -> decidable Halt` as notion of undecidability in the library.", so I used this definition.

DmxLarchey: Yep you are right. But the remark still holds. What make you think `decidable P -> decidable Halt`  with `decidable P <-> inhabited (dec P)`, is a good _characterization_ of undecidability for `P`. For me it is just a stronger notion than "meta-level undecidable". Much like `decidable` is a stronger notion than meta-level decidable.

yforster:
I disagree on that the fact that we use synthetic definitions should be made more clear in the names. The definition of decidable for instance is in the file `Undecidability.Synthetic.Definitions`, I don't think that will confuse people. But again, that's orthogonal to the PR.

> you said that while you agree they can be used to show undecidability results by reduction (because they are stronger than non-informative reduction), using them to characterize undecidability leads to statements contradicting regular computability theory.

No, that's not the issue, I probably was not clear enough.

We characterise undecidability as `decidable p -> decidable Halt`, as agreed in the call. Neither informative nor non-informative reductions play a role in the definition.

What I said is that using informative definitions everywhere leads to theorems contradicting traditional computability theory.

For instance one can prove `forall p : nat -> X -> Prop, (forall n, inf_enumerable (p n)) -> inf_enumerable (fun x => exists n, p n x)`, and I think you even proved something similar in the Trakhtenbrot paper. The statement that the countable union of enumerable sets is enumerable is however *false* in traditional computability theory, so taking `inf_enumerable` as our definition of enumerability would not be faithful.

> Fair enough but what makes you think this would not also happen with non-informative synthetic undecidability ?

Undecidability is defined as a logical implication, so I don't see how one could define this informatively.

But more generally, I believe that the non-informative definitions do not allow one to prove theorems contradicting traditional computability. This is because lots of research on constructive reverse mathematics suggests that CT seems to be inconsistent with classical logic only under the presence of choice axioms. Since I believe (it's really a believe) that Coq can not prove the required choice axioms I think that CT is consistent in Coq with classical logic as used in traditional computability - and under CT our non-informative definition of enumerability can be shown equivalent to the traditional definition involving models of computation in Coq.

I am working on putting together all the relevant references to the literature and explain them in Coq's setting rather than something like HA, but that takes time :)

theories/CFG/CFP_undec.v
yforster: It's the exact definition we published in the CPP '19 paper on BSMs, MMs, and ILL (back then named `undec` instead of `undecidable`) and we all agreed in the Zoom call that we should use it as the central definition in the library. You gave a long and detailed explanation why that's a good definition at CPP '19, I gave a similar and even longer explanation at CoqPL '20. Nobody at CPP, at CoqPL, or in the Zoom call suggested any reason why that's not a good definition. It's furthermore equivalent to `~ decidable P` given the "Church's thesis" axiom we discussed, which is proven consistent for univalent type theory in https://arxiv.org/abs/1905.03014.

DmxLarchey:
> Undecidability is defined as a logical implication, so I don't see how one could define this informatively.

Just a quick remark about how this could be done:

```
inf_dec (P : X -> Prop) := forall x : X, { P x } + { ~ P x }.
inf_undec P := inf_dec P -> inf_dec Halt.
```

In this case, the "implication" is in reality a _map_ that computationally transforms a decider (ie. `inf_dec`) for `P` into one for `Halt`.

mrhaandi:
I'm on the same page as Yannick and approve of his design of the PR. I am not in favor of an additional `inf_undec`.

theories/CFG/CFP_undec.v
DmxLarchey: What I am saying that `decidable P -> decidable Halt` it is a good definition for synthetic undecidability. Do not get me wrong, I am not questioning that.

But I have the feeling, may be I am wrong, that there is a push to identify synthetic undecidability with undecidability (which is not clearly defined and reader dependent) and this I think would be a wrong move. In what I understood from Y. Bertot's remarks at CoqPL, I do think he believed we did prove undecidability wrt. a concrete model of computation. Maybe this holds for some of our problems wrt. `L`-defined concrete undecidability.

Concerning models, it would indeed be good if we could say that there is a model where CT holds and that we can use any axiom satisfied within that model.

yforster:
The point of defining `undecidable` is to have a nice and compact undecidability theorems that we can point users to. The technically interesting parts are always somewhere else, it's just about picking a definition which is philosophically nice that we want to expose to users. I don't mind contributors using arbitrary auxiliary definitions to come to their result, and if someone wants to use `inf_undec` as auxiliary definition that's okay.

But that's not the definition of undecidability we should choose as our formal definition of the concept and expose to users.

We know that `inf_enumerable` is not the right definition of enumerability to expose to users. Thus, `enumerable` is the formal definition of enumerability we want to use and expose. But then `inf_decidable` can't be the right definition of decidability we want to expose, because it doesn't correspond to the chosen definition of enumerability. So `decidable` is the formal definition of decidability that we want to use and expose.

Now it would be very strange to pick `inf_undec` as our formal definition of undecidability. It is defined using notions we do not want people to think about because they lead to results contradicting traditional computability theory.

theories/CFG/CFP_undec.v
DmxLarchey: By CT, I mean of course that the type `nat -> nat` is exactly interpreted as the collection of total computable functions from N to N in the model.

DmxLarchey:
May be I am misleading you both in this discussion. I am **NOT** pushing for informative definitions as our main definitions. I replied to the point you made by saying that `->`can be viewed as a logical implication, but also as a computable map. 

What I would like to refrain from is for people to discover one day, as Y. Bertot did seemingly, that when we state `undec P`, this does not mean _exactly_ there is no TM for deciding `P`, it only implies it when the axioms which are used to prove  `undec P` do not introduce non-computable functions in Coq. 

We do not cheat about that in the papers I think, but people may have pre-conceptions in their minds that might induce them into believing whatever they are used to when we use a well-established terminology.

DmxLarchey:
> For instance one can prove `forall p : nat -> X -> Prop, (forall n, inf_enumerable (p n)) -> inf_enumerable (fun x => exists n, p n x)`, and I think you even proved something similar in the Trakhtenbrot paper. The statement that the countable union of enumerable sets is enumerable is however _false_ in traditional computability theory, so taking `inf_enumerable` as our definition of enumerability would not be faithful.

Thanks for this point, it is clear now what you mean for much too strong `inf_enumerable`...

DmxLarchey:
@yforster: back to when we were designing the CPP'19 paper together, I remember a proposition you had about adding the rule `undecidable' P -> undecidable' ~P` to the inductive characterization that we proposed there. I do not remember exactly why but this had something to do with other developments related to FOL (and maybe Markov ...)

Now to both @yf and @mrhaandi, if you consider the following development, you see that `undecidable' P` with negation (3 rules) is weaker than regular `undecidable P` (with 2 rules and equivalent to `decidable P -> decidable Halt`). However, when adding eXcluded Middle `XM`, these two notions become equivalent. So we certainly cannot refute that the `undecidable` without negation is strictly stronger than `undecidable'` but I doubt we can prove the equivalence in general, ie w/o `XM`.

So the question is whether a rule such as eg. closure of `undecidable` under negation should be allowed in the synthetic framework or not ? My point is, should we _freeze_ a definition of `undecidable`.

Or instead, take another approach which would be to axiomatize `undecidable` with _closure rules_ that we allow and that could possibly be extended? Hence, we would have the constructors but contrary to the inductive definition, we would not have the eliminator `undecidable_ind`. 

Letting the definition of `undecidable` _open_ would still allow deriving undecidability results by eg. reduction but would mostly forbid meta-reasoning on undecidability.   

```
Require Setoid.

Set Implicit Arguments.

Definition dec X (P : X -> Prop) := forall x, { P x } + { ~ P x }.
Definition decidable X P := inhabited (@dec X P).

Definition XM := forall (P : Prop), P \/ ~ P. 

Fact dec_sigma X P : (dec P -> { d : X -> bool | forall x, P x <-> d x = true }) 
                   * ({ d : X -> bool | forall x, P x <-> d x = true } -> dec P).
Proof.
  split.
  + intros f; exists (fun x => if f x then true else false).
    intros x; destruct (f x) as [ H | H ]; split; try tauto; discriminate.
  + intros (d & Hd) x.
    case_eq (d x); intros Hx.
    * left; rewrite Hd; auto.
    * right; rewrite Hd, Hx; discriminate.
Qed.

Fact decidable_exists X P : decidable P <-> exists d, forall x : X, P x <-> d x = true.
Proof.
  split.
  + intros [ H ]; apply dec_sigma in H; destruct H as (d & Hd); exists d; auto.
  + intros (d & Hd); exists; apply dec_sigma; exists d; auto.
Qed.

Fact neg_dec X P : @dec X P -> dec (fun x => ~ P x).
Proof.
  intros d x; generalize (d x); tauto.
Qed.

Fact dec_neg (c : XM) X P : dec (fun x : X => ~ P x) -> dec P.
Proof.
  intros H1 x.
  destruct (H1 x); try tauto.
  left; generalize (c (P x)); tauto.
Qed.

Fact neg_decidable X P : @decidable X P -> decidable (fun x => ~ P x).
Proof.
  intros [ H ]; exists; revert H; apply neg_dec.
Qed.

Fact decidable_neg (c : XM) X P : decidable (fun x : X => ~ P x) -> decidable P.
Proof.
  intros [ H1 ]; exists; revert c X P H1; apply dec_neg.
Qed.

Section undecidable.

  Variables (TM : Type) (Halt : TM -> Prop).

  Inductive undecidable : forall X, (X -> Prop) -> Prop :=
    | undecidable_seed : undecidable Halt
    | undecidable_tred : forall X Y P Q, (@decidable Y Q -> @decidable X P) 
                                       -> undecidable P 
                                       -> undecidable Q.

  Fact undecidable_equiv X P : @undecidable X P <-> (decidable P -> decidable Halt).
  Proof.
    split.
    * induction 1; auto.
    * intros H; constructor 2 with (1 := H); constructor.
  Qed.

  Inductive undecidable' : forall X, (X -> Prop) -> Prop :=
    | undecidable'_seed : undecidable' Halt
    | undecidable'_neg  : forall X P, undecidable' P -> undecidable' (fun x : X => ~ P x)
    | undecidable'_tred : forall X Y P Q, (@decidable Y Q -> @decidable X P) 
                                        -> undecidable' P 
                                        -> undecidable' Q.

  Fact undecidable_undecidable' X P : @undecidable X P -> undecidable' P.
  Proof.
    induction 1 as [ | X Y P Q H ].
    + constructor.
    + constructor 3 with (1 := H); auto.
  Qed.

  Fact undecidable'_undecidable (c : XM) X P : @undecidable X P <-> undecidable' P.
  Proof.
    split.
    + apply undecidable_undecidable'.
    + induction 1 as [ | X P H IH | X Y P Q H1 H2 IH2 ].
      * constructor.
      * constructor 2 with (2 := IH); apply decidable_neg; auto.
      * constructor 2 with (1 := H1); auto.
  Qed.

End undecidable.
```

theories/CFG/CFP_undec.v
DmxLarchey: Could CT be expressed as:

```
CT : forall f : nat -> nat, exists a : recalg 1, forall x y, y = f x <-> [|a|] <x> y
```

meaning every Coq definable term `nat -> nat` is extensionally the computation of a µ-recursive algorithm of arity 1.

Recall that

```
forall a : recalg 1, (forall x, exists y, [|a|] <x> y) 
                  -> { f : nat -> nat | forall x, [|a|] <x> (f x) }.
```

is provable in axiom free Coq, as witnessed in my ITP'17 paper, using Constructive Epsilon.

yforster:
I was indeed mislead (or probably, mislead myself) :) I think we're mostly on the same page!

I don't have strong opinions on whether or not we should freeze a definition of undecidability. I used to have the strong opinion that we *should not* freeze one. After giving talks and explaining the topics to outsiders my view changed: To introduce the concept, it's good to talk about a fixed definition, so we might as well put that definition into the code.

In the call I had the impression that people are happy fixing a definition, so I was happy to do so, but that can be discussed. 

My proposal would be to take this as a reason to schedule another Zoom call - it can probably be short. At least for myself the call gave me lots of motivation to look at more annoying parts of the library, maybe we see this effect again.

DmxLarchey:
Hi @yforster. I think indeed a short Zoom session would be helpful. As you, I do not think we are at odds. But I prefer that we remain very precise about what we do prove and what we believe to be true. Writing this library, we got used into identifying, at least in the discourse, synthetic Coq computability notions (decidability, computability, undecidability) with the "real" notions. For me, synthetic notions do not make sense w/o a normalization theorem, hence they interact with axioms in a subtle way. Remember that we claimed that we have the computability of Coq-reductions for free **because** we work with Coq and not with eg HOL or Isabelle/HOL. This computability of course fails with inconsistent axioms; but consistency is not enough.

yforster:
@DmxLarchey are you fine with me merging this PR? It would help Andrej to not have diverging developments, and especially since the files in which things are proven change that's pretty relevant. (If we decide later on that we want to name things differently we can still do that.)

PR #51, opened 2020-04-30
Rebased Hoare framework to 8.11

mwuttke97:
Rebase of PR #19. Compiles fine with 8.11.


mwuttke97:
The commit is so big, because (back in October) I copied the entire framework code from the private repo into this repo and added `From Undecidable` in all imports.

Also note that this code can't compile on older versions (pre 8.9 IIRC) due to a bug in Coq.

yforster:
Does this subsume our CPP '19 development?

mwuttke97:
Yes. IIRC, this work was done after the CPP'20 development. In addition, it has the following features (besides Hoare):
- Rename `isRight` into `isVoid`. Also in code that uses the old name
- Remove Haskell extraction
- A fix for the tapes lift and additional stronger correctness lemmas for operators (needed only for Hoare right now)

It may still needs some cleanup:
[ ] Fix whitespaces in all files (e.g. using `M-x whi-c` in all files)
[ ] Remove `Check`s and `Compute`s in the new code.

yforster:
Do we get faster compilation for the translation L -> TM or the mTM -> sTM compiler like this, or are this parts untouched?

mwuttke97:
These parts are untouched as of now. However, there is a *seperate* verification of `Univ` and some list machines. Last autumn I started porting the L interpreter to Hoare, but I haven't continued working there. The `mTM -> sTM` compiler is out of scope for the Hoare framework, also. Basically, this framework was meant to make future developments easier (e.g. the POPL paper).

Also, I just removed an unused (and long obsolete) tactic which was still called in `TMSimp`, which could speed up `TMSimp` slightly.

mwuttke97:
These parts are untouched as of now. However, there is a seperate verification of Univ (which compiles very fast) and some list machines. Last autumn I started porting the L interpreter to Hoare, but I haven't continued working there. The mTM -> sTM compiler is out of scope for the Hoare framework, also. Basically, this framework was meant to make future developments easier (e.g. the POPL paper).

Also, I just removed an unused (and long obsolete) tactic which was still called in TMSimp, which could speed up TMSimp slightly.

yforster:
So at the moment both the old verification of Univ and the new verification are included? I think then we should just remove the old verification and just keep the new one here. As I understand that also should improve the compilation time of the library?

In the (near? @fakusb) future we should move all things which morally belong to the complexity layer of our work to a separate repository maintained by Fabian. 

If we could get a faster L interpreter using the Hoare framework that would be amazing, but I'm happy to merge also without it after we remove unnecessary code like the old Univ verification.

Also, the library is warning and printing free at the moment, it would be great if we can keep it like this :)

mwuttke97:
>> So at the moment both the old verification of Univ and the new verification are included? I think then we should just remove the old verification and just keep the new one here.

I could merge the two Univ verifications if I have a bit spare time. The low-level part of Univ (i.e. operations on the "object tape") is the only part that is not (and is not supposed to) be done in the Hoare framework. Likewise, the M2M compiler is not meant at all to be verified with Hoare. I haven't commited any code for the L re-verification, but I guess that would be not to difficult. I guess that the speedup for L will be significant. (The bottleneck were the tape rewrites, which is exactly what is *not* done in Hoare.)

>> As I understand that also should improve the compilation time of the library?

If I recall correctly, the verification of the high-level parts of Univ was much faster with Hoare, but Univ didn't take long time to compile anyway. I am not sure whether it will boost the overall compile time of the existing library, but it will definitely boost compilation time after L has been reverified.

>> In the (near? @fakusb) future we should move all things which morally belong to the complexity layer of our work to a separate repository maintained by Fabian.

That makes sense. In this repo we don't really need the running time specifications. Is this also true for memory usage? We could have a simplified version of 'termination relations' that don't account for the actual running time. We could also simplify the termination Hoare triples by removing the step counts.

mwuttke97:
> So at the moment both the old verification of Univ and the new verification are included? I think then we should just remove the old verification and just keep the new one here.

I could merge the two Univ verifications if I have a bit spare time. The low-level part of Univ (i.e. operations on the "object tape") is the only part that is not (and is not supposed to) be done in the Hoare framework. Likewise, the M2M compiler is not meant at all to be verified with Hoare. I haven't commited any code for the L re-verification, but I guess that would be not to difficult. I guess that the speedup for L will be significant. (The bottleneck were the tape rewrites, which is exactly what is *not* done in Hoare.)

> As I understand that also should improve the compilation time of the library?

If I recall correctly, the verification of the high-level parts of Univ was much faster with Hoare, but Univ didn't take long time to compile anyway. I am not sure whether it will boost the overall compile time of the existing library, but it will definitely boost compilation time after L has been reverified.

> In the (near? @fakusb) future we should move all things which morally belong to the complexity layer of our work to a separate repository maintained by Fabian.

That makes sense. In this repo we don't really need the running time specifications. Is this also true for memory usage? We could have a simplified version of 'termination relations' that don't account for the actual running time. We could also simplify the termination Hoare triples by removing the step counts.

mwuttke97:
> So at the moment both the old verification of Univ and the new verification are included? I think then we should just remove the old verification and just keep the new one here.

I could merge the two Univ verifications if I have a bit spare time. The low-level part of Univ (i.e. operations on the "object tape") is the only part that is not (and is not supposed to) be done in the Hoare framework. Likewise, the M2M compiler is not meant at all to be verified with Hoare. I haven't commited any code for the L re-verification, but I guess that would be not too difficult. I guess that the speedup for L will be significant. (The bottleneck were the tape rewrites, which is exactly what is *not* done in Hoare.)

> As I understand that also should improve the compilation time of the library?

If I recall correctly, the verification of the high-level parts of Univ was much faster with Hoare, but Univ didn't take long time to compile anyway. I am not sure whether it will boost the overall compile time of the existing library, but it will definitely boost compilation time after L has been reverified.

> In the (near? @fakusb) future we should move all things which morally belong to the complexity layer of our work to a separate repository maintained by Fabian.

That makes sense. In this repo we don't really need the running time specifications. Is this also true for memory usage? We could have a simplified version of 'termination relations' that don't account for the actual running time. We could also simplify the termination Hoare triples by removing the step counts.

yforster:
Superseded by #72

PR #52, opened 2020-06-28
Semi-unification (and more)

mrhaandi:
- added undecidability of the following problems
  + semi-unification
  + uniform H10 constraint solvability
  + one counter machine halting
  + uniform boundedness of stack machines
  + problems for Hilbert-calculi
  + linear polynomial constraint solvability
  + finite multiset constraint solvability
- moved SR to StringRewriting and used new synthetic definitions
- moved elementary H10 constraint solvability to DiophantineConstraints


yforster:
The Travis run is here, no idea why it doesn't show up properly in the PR: https://travis-ci.org/github/uds-psl/coq-library-undecidability/builds/702937446

DmxLarchey:
Hi Andrej,

Looking at the diffs, there are many repetitions for some utility libraries like eg `measure_ind` but also list libraries etc ? Don't you want to create you own shared lib with those ?

Best,

D.

mrhaandi:
@DmxLarchey 
> Looking at the diffs, there are many repetitions for some utility libraries like eg `measure_ind` but also list libraries etc ? Don't you want to create you own shared lib with those ?

I want individual undecidability proofs (especially orthogonal ones) to be as modular as possible.
Since the overlap is less than 10% of the overall proofs and can be compiled in parallel, it does not introduce too much burden. I also pay attention to each time include only the necessary lemmas (not just copy-pasting).

The only downside would be that if some future Coq version breaks a lemma, then it will need a fix at several occurrences.
On the other hand, if the auxiliary lemmas were centralized, then any modification of the shared library would require adjusting older proofs again.
Also, a shared library design requires a lot of forethought and software engineering to do right.

Personally, I felt the negative aspects of ad hoc shared libraries while interacting with shared.prelim and also the the uds library.

DmxLarchey:
I recognized that there are some downsides but my experience is that they are not so important in practice. You can stick to your scheme but may I suggest that you maintain a shared lib with the most recent version of your lemmas and proofs, even if you do not `Import` them. This would maintain a faithful trunc for your development.

PR #53, opened 2020-07-17
singleTM and string rewriting dependencies cleanup

mrhaandi:
- cleaned up singleTM dependencies
- cleaned up string rewriting dependencies
- added many-one reductions for semi-unification
- added functional two-counter machines


DmxLarchey:
What is the difference between CM2 and 2 counters Minsky machines ?

DmxLarchey:
Another difference is the semantics. Here (`CM2`), instructions always succeed, `step` is a total function but termination is witnessed by the state being stalled. Any regular instruction must change either the `PC` or else decrease one of the registers strictly. This is also true for `MM2`, however the step relation fails on _empty_ instructions. But not true for `MM(n)` because the semantics of `DEC` is different. There (`MM(n)`) programs can stall while executing `i : DEC x i` if `x` has value `0`. SO the termination condition for `MM(n)` is not equivalent.

DmxLarchey:
Yannick and Andrej,

Do you think we should create a directory for counter machines and another for stack machines. Right now, the code for MM and BSM is in the ILL directory, not ideal ... On the contrary, MuRec is toplevel. MM are used in both ILL, H10 and the reduction MM(n) -> FRACTRAN -> MMa(2)

BSM, I do not know how similar they are to the stack machines involved in the semi-unification proof.

D.

mrhaandi:
> Do you think we should create a directory for counter machines and another for stack machines. 

Stack machines and counter machines are often good _seeds_ and deserve their own top level folder (as it is currently the case).

> Right now, the code for MM and BSM is in the ILL directory, not ideal

This is definitely not ideal and should be refactored to follow the current library design guidelines.

> BSM, I do not know how similar they are to the stack machines involved in the semi-unification proof.

BSM is has separate push/pop instructions, which do not preserve the sum of lengths of the stacks.
SSM on the other hand always pops and pushes a symbol simultaneously, thus operating in bounded space.

theories/H10/H10.v
```diff
@@ -58,35 +55,3 @@ Proof.
...
     rewrite !dp_inst_par_eval; auto.
 Qed.
-
-Theorem Fractran_UNDEC : Halt ⪯ FRACTRAN_HALTING.
-Proof.
-  apply reduces_transitive with (1 := MM_HALTING_undec).
-  exact MM_FRACTRAN_HALTING.
-Qed.
-
-Theorem Hilberts_Tenth : Halt ⪯ PCP
```
DmxLarchey, line 68: This theorem corresponds to the statement of the reduction chain in the H10 FSCD paper, I would like to keep it somewhere.

mrhaandi: > Halt ⪯ PCP
> I would like to keep it somewhere

is now ``Lemma HaltTM_to_PCP : HaltTM 1 ⪯ PCP``
and part of ``theories/PCP/Reductions/HaltTM_to_PCP.v`` as it is inherent to PCP and not H10.
This lemma is still used in the proof of ``Lemma HaltTM_to_H10 : HaltTM 1 ⪯ H10`` in ``theories/H10/HaltTM_to_H10.v``
By the way, ``Halt`` was deprecated in favor of ``HaltTM 1`` in the formulation of synthetic reductions.

DmxLarchey: No I think you are missing the point: `Hilberts_Tenth` is **not** `Halt ⪯ PCP` but the conjunction of the whole chain (see below the messages (lines 69--74). 

I wanted that all the steps in the chain to appear in the _statement_ of `Hilberts_Tenth`. Of course they do appear in the proof.

mrhaandi: I see, you want ``Halt ⪯ PCP /\ PCP ⪯ MM_HALTING /\ MM_HALTING ⪯ FRACTRAN_HALTING /\ FRACTRAN_HALTING ⪯ DIO_LOGIC_SAT ...`` to be a lemma somewhere (I did not understand the quotation).
It could be put into ``HaltTM_to_H10.v`` since it has similar dependencies. Do you agree?
Also, the H10 folder still will need refactoring to follow the current design guidelines.

mrhaandi: I now added ``Theorem Hilberts_Tenth`` to ``HaltTM_to_H10.v`` with the reference to the FSCD 2019 paper.

yforster:
As far as I understand, all questions here are resolved (or actively discussed in #54), so I'll merge in a moment to make it easier for #54 not to diverge too much.

DmxLarchey:
Ok you can merge, I will try to sync. it with #54 later on ...

PR #54, opened 2020-07-23
Restructured Minsky machines, Binary Stack machines and FRACTRAN

DmxLarchey:
The following sub-trees have been but at top-level:

- `MinskyMachines`
- `BinaryStackMachines`
- `FRACTRAN`

The generic low-level code compiler (CPP'19) was moved from `ILL/Code` to `Shared/Libs/DLW/Code`

The project should compile, it does on my computer but `CI` seems down ?


DmxLarchey:
Hi Yannick,

This is still `WIP` (work in progress) but nearly finished.

Is `CI` disabled ?

yforster:
I don't understand why CI doesn't show up anymore in the PR overview, but it's still running :) All your builds went fine: https://travis-ci.org/github/uds-psl/coq-library-undecidability/builds/711279882

DmxLarchey:
Ok thank you. Well then, if you both agree with the new structure, we could merge.

mrhaandi:
Binary stack machines should go into the ``StackMachines`` folder (it is a stack machine problem after all) and follow the previously discussed design guidelines:
- StackMachines/BSM.v contains minimal problem definitions (no lemmas etc.)
- StackMachines/BSM_undec.v contains the few lines proof that the problem is ``undecidable`` as per synthetic definitions
- StackMachines/Reductions may contain e.g. the reduction from PCP to BSM_HALTING
- all auxiliary infrastructure has to go into subdirectories, e.g. StackMachines/Util or StackMachines/BSM/Util, and not be at the same level as problem definition and the undecidability statement

Same design guidelines apply to Minsky machines and FRACTRAN.
Minsky machines could also be put into the ``CounterMachines`` folder as Minsky machines are a subclass of Counter machines.

yforster:
I agree with Andrej in the sense that this should be the final state of the library. So if you're willing to do that @DmxLarchey, great.

But my personal opinion is that we can also accept PRs which are a step towards the final state of the library, even if they do not follow the guidelines (yet).

That's up for discussion though. What do you think?

I agree that `BSM` would fit well into `StackMachines`. Concerning `MM` I'm not so sure, but I would not be against moving them to `CounterMachines`.

DmxLarchey:
Ok I did make the effort (and it took some time) to reorganise the following directories according to the new guidelines: `BinaryStackMachines`, `MinskyMachines` and `FRACTRAN`

I also moved `Reductions` in those directories according to the target of the reduction, maintaining a Coq symbolic link into the top-level (legacy) `Reductions` directory.

Before I do merge e.g. `BinaryStackMachines` into `CounterMachines`, I would like Andrej to check whether conforms the new strict guidelines ... Notice that most of the `Undecidability` library *does not* follow those guidelines ATM.

Concerning Minsky machines, it is a complicated case because they are somehow central and many pointers in & out. Plus there is a cycle: `MM` -> `FRACTRAN` -> `MMA2` -> `MM2`, meaning that the directories `MinskyMachines` and `FRACTRAN` cannot be compiled in a sequence. 

Obviously, `MM2` would be the preferred `seed` but to get the `undec` for this seed, you go through the above cycle. Also, when `MM` is a target, if the user just uses the definition, then he will end up rewriting most of the `Code/sss.v` and `MM/mm_defs.v` libraries anyway.

So `MinskyMachines` is a complicated case right now.

DmxLarchey:
> I agree that `BSM` would fit well into `StackMachines`. Concerning `MM` I'm not so sure, but I would not be against moving them to `CounterMachines`.

I can move `BinaryStackMachines/*` into `StackMachines` once you agree that the structure in now in conformity. 

Concerning Minsky machines, we need to have a discussion because it is a very connected set of nodes, see my comment.

DmxLarchey:
Now `ILL` also should conform the guidelines ... introduction of the notion of _reduction chain_ to display the results as they exist in papers like e.g. `CPP'19` and `FSCD'19` .

mrhaandi:
I can confirm that ``BinaryStackMachines`` follows the current design guidelines.
Minor things that I noticed:
- ``sss_terminates`` is used in problem specification, which took me some time to understand looking at the shared library.
For example, it requires to understand ``sss_output P st st' := P // st ->> st' /\ out_code (fst st') P.``
It would help in ``Code/sss.v`` lines 17-24 (the ``Reserved Notation`` block) to mention for each line as a comment the intended meaning of the notation. I liked the ``MM2.v`` in that regard, which I was able to give to students as is.
- the content of ``bsm_pcp.v`` may go into ``Reductions/iBPCP_BSM.v``, as it is specific for this particular reduction and is neither used for parallel compilation (as individual file) nor anywhere else as Import.
- the file ``iBPCP_BSM.v`` may be named ``iPCPb_to_BSM_HALTING`` using the actual problem names. There are several reasons.
  + if at a later point another problem for binary stack machines is added, it gets confusing
  + some reductions actually go both ways, see ``PCP/Reductions/PCPb_iff_dPCPb.v``, so ``_to_`` signifies the direction
  + looking at the filename the main theorem is more apparent (better for search)

I also looked at ``MinskyMachines`` and similar observations as above apply.
Overall, the current refactoring is a good improvement and can be merged anytime, as the above observations are only minor ones.

mrhaandi:
> meaning that the directories ``MinskyMachines`` and ``FRACTRAN`` cannot be compiled in a sequence

The important point is that the individual (atomic) reduction steps can be compiled in parallel, as it takes the most time.
The final ``_undec`` lemmas may be compiled in a final, sequential step. This should be almost instantaneous, as the proofs of the ``_undec`` lemmas should be simple compositions of the atomic reduction steps.

> introduction of the notion of reduction chain

I agree that it is very good to have paper-related reduction chain lemmas (such as ``Hilberts_Tenth``), which in the theorem statement show the high-level reduction chain.
However, I'm not sure whether the extra notations introduced in ``H10_undec.v`` by this PR are the most clear way to display them. 
To me ``HaltTM 1 ⪯ X /\ X ⪯ Y /\ Y ⪯ Z`` looks fine enough.
On paper, one would want to write ``HaltTM 1 ⪯ X ⪯ Y ⪯ Z``, but I'm not sure whether Coq notations could be used for that.
Also ``⪯ₗ`` looks like a one-one reduction, as opposed to many-one.

DmxLarchey:
Merged with the current `coq-8.11` branch after PR #53 and #55

DmxLarchey:
>     * It would help in `Code/sss.v` lines 17-24 (the `Reserved Notation` block) to mention for each line as a comment the intended meaning of the notation.

Done

DmxLarchey:
>     * I liked the `MM2.v` in that regard, which I was able to give to students as is.

Sure but this only works well when `MM2` is a _source_. Now think of it as a _target_, for instance, I suggest you try to implement the reduction `MM2_HALTING ⪯ MM2_HALTS_ON_ZERO` to get a glimpse of why the short description of the semantics given for `MM2` does not work well when `MM2` is a target.

The problem with `MM` is that they are both a source (eg `MM ⪯ FRACTRAN`) and a target  (eg `BSM ⪯ MM` or `FRACTRAN ⪯ MMA2`). You can also look at `MMA2 ⪯ MM2` which is overly complicated for just the identity map !!

DmxLarchey:
>     * content of `bsm_pcp.v` may go into `Reductions/iBPCP_BSM.v`

This might hinder parallel compilation because `bsm_pcp.v` does not depend on the definitions in `PCP` while `iBPCP_BSM.v` does. Btw `tiles_solvable` was defined way before we actually had a naming scheme/convention for `PCP` problems.

DmxLarchey:
>     * may be named `iPCPb_to_BSM_HALTING`

Done. Maybe each time you spot such an issue, tell me.

DmxLarchey:
> However, I'm not sure whether the extra notations introduced in `H10_undec.v` by this PR are the most clear way to display them.

Ok, that if you have a better solution, I am open to other choices.

> To me `HaltTM 1 ⪯ X /\ X ⪯ Y /\ Y ⪯ Z` looks fine enough.

Well the thing is you need to repeat the intermediate steps (ie `X` and `Y` over there) twice ...

> On paper, one would want to write `HaltTM 1 ⪯ X ⪯ Y ⪯ Z`, but I'm not sure whether Coq notations could be used for that.

That was my problem indeed...

> Also `⪯ₗ` looks like a one-one reduction, as opposed to many-one.

Ok the `l` subscript (for list) is not ideal as it can be confused with a `1`. Not every subscript exists in unicode though... I will try something else.

mrhaandi:
> Sure but this only works well when `MM2` is a _source_. 

I agree. The main design principle is that if something is a _seed_, i.e. intended to be used as a source often, then there has to be maximal emphasis on a minimalistic definition. That is why I think that for `MM2` it is very well done.

If something is a final sink, e.g. semi-unification, a minimalistic definition helps an outside reader to understand that the code actually mechanizes the underlying problem (e.g. in a published paper).

For intermediate problems such as BSM it is not that important. I personally still prefer to completely separate 
- minimalistic problem definition (show to students, paper readers, etc.)
- reusable, reduction independent problem infrastructure (reuse to argue about the problem in reductions from/to; show to co-developers)
- atomic reductions (compiled in parallel after definitions and infrastructure; no need to look at the code working on other projects)
- final undecidability statement (show to students, paper readers, etc.)

> This might hinder parallel compilation because `bsm_pcp.v` does not depend on the definitions in `PCP` while `iBPCP_BSM.v`. 

I see. The name `bsm_pcp.v` to me suggested that it uses `PCP`. Then it is better to leave as is.

DmxLarchey:
> If something is a final sink, e.g. semi-unification, a minimalistic definition helps an outside reader to understand that the code actually mechanizes the underlying problem (e.g. in a published paper).

I agree this is important, and actually a problem for the definition of Turing machines in this library. However, is there a short and self-contained way to describe what is `HaltTM` ?

mrhaandi:
> However, is there a short and self-contained way to describe what is `HaltTM`?

@yforster is currently working at exactly that.

>> To me HaltTM 1 ⪯ X /\ X ⪯ Y /\ Y ⪯ Z looks fine enough.

> Well the thing is you need to repeat the intermediate steps (ie X and Y over there) twice ...

I discussed this @yforster, and we both prefer repeating the intermediate steps compared to introducing new notations.
For both Coq developers and mathematicians this should pose no problem.
If your main concern is to save a few lines in a paper quoting code (which is somewhat orthogonal to library design), then maybe [notations-with-recursive-patterns](https://coq.github.io/doc/v8.12/refman/user-extensions/syntax-extensions.html#notations-with-recursive-patterns) could be utilized.

DmxLarchey:
`BSM`, `MM`, `MMA` and `MMA2` share the `Code/sss` infrastructure because of:

+ the compiler that is used several time;
+ also, somehow complicated programs are written in these, and good tools to reason with them help a lot.

You can ask @yforster who wrote the `MM_no_loops` to `FRACTRAN` reduction (w/o knowledge of the above `sss` infrastructure) what is his experience of programming low-level code. For instance, if you do not use the _determinism_ of the target model, you end up writing some proofs twice: on the way down and on the way up.

Even if `BSM`, `MM`, `MMA` and `MMA2` were _seeds_ at some point, I do not think they are any more. `MM2` is a good replacement for `MM`. Now we have both "termination" and "termination on zero" for `MM2`. That should suffice as seeds. 

`BSM` was never intended to be a seed but I could give a simplified `BSM2` version. However, an easy reduction from `MM2` should do. But again, it is much easier with the compiler, ie via `MMA2`. 

Personality, I see no reason to use `BSM2` when `MM2` is available, `MM2` being just `BSM2` with _unary_ stacks instead of _binary/Boolean_ stacks. 

`MM1` is decidable (from _Bouajjani, Esparza & Maler_ I think), for `BSM1` I do not know.

mrhaandi:
> Even if BSM, MM, MMA and MMA2 were seeds at some point, I do not think they are any more. MM2 is a good replacement for MM.

I agree. No need to change that.

> Now we have both "termination" and "termination on zero" for MM2. That should suffice as seeds.

A common seed that I know (that is used for semi-unification) is termination with initial counters (0, 0), as opposed to the counter values (a, b) being part of the input.

> I see no reason to use `BSM2` when `MM2` is available.

I agree.

> for `BSM1` I do not know.

This feels like a (one-stack) push-down automaton, which should be context free (decidable).

DmxLarchey:
> A common seed that I know (that is used for semi-unification) is termination with initial counters (0, 0), as opposed to the counter values (a, b) being part of the input.

Have you implemented it? It not, it should not be difficult to add the encoding the counters initialization into the programs.

mrhaandi:
> Have you implemented it?

It is inherent in `CounterMachines` to both

```
(** Two Counter Machine Halting Problem *)
Definition CM2_HALT : Cm2 -> Prop :=
  fun M => exists (n: nat), 
    halting M (Nat.iter n (step M) {| state := 0; value1 := 0; value2 := 0 |}).
```
and
```
(** One Counter Machine Halting Problem *)
Definition CM1_HALT : Cm1 -> Prop :=
  fun M => exists (n: nat), 
    halting M (Nat.iter n (step M) {| state := 0; value := 1 |}).
```
that I used for student work and semi-unification.

DmxLarchey:
@yforster and @mrhaandi,

I think there is enough changes in here. I would like it to be merged. Further cleanups can come later.

mrhaandi:
> I think there is enough changes in here. I would like it to be merged. Further cleanups can come later.

I agree.

DmxLarchey:
> maybe [notations-with-recursive-patterns](https://coq.github.io/doc/v8.12/refman/user-extensions/syntax-extensions.html#notations-with-recursive-patterns) could be utilized

Thanks to M. Wuttke via `coq-club`, we have a nice notation for reduction chains now, IMHO, eg

```
Theorem Hilberts_Tenth : 
  ⎩ HaltTM 1 ⪯ₘ PCPb ⪯ₘ MM_HALTING ⪯ₘ FRACTRAN_HALTING ⪯ₘ DIO_LOGIC_SAT ⪯ₘ DIO_ELEM_SAT ⪯ₘ DIO_SINGLE_SAT ⪯ₘ H10 ⎭.
```

mrhaandi:
I suggest this PR to be merged soon, before further changes.
The recent changes to the synthetic definition files along with new notations and tactics are orthogonal to the previous refactoring and should be discussed in a separate PR.

Although an improvement over the previous suggestion, I still would not advise to use the current reduction chain notation.
Just to understand the statement
```
Theorem Hilberts_Tenth : 
  ⎩ HaltTM 1 ⪯ₘ PCPb ⪯ₘ MM_HALTING ⪯ₘ FRACTRAN_HALTING ⪯ₘ DIO_LOGIC_SAT ⪯ₘ DIO_ELEM_SAT ⪯ₘ DIO_SINGLE_SAT ⪯ₘ H10 ⎭.
```
a student/reader would require to understand
- `ltac:_` terms
- the recursive `redchain2Prop` tactic
- `lazymatch`
- `constr:_`
- `?x`
To be honest (since I'm not too experienced with ltac), I have an intuition, but do not fully understand what is going on in the background just to formulate the reduction chain statement.

DmxLarchey:
> Just to understand the statement
> 
> ```
> Theorem Hilberts_Tenth : 
>   ⎩ HaltTM 1 ⪯ₘ PCPb ⪯ₘ MM_HALTING ⪯ₘ FRACTRAN_HALTING ⪯ₘ DIO_LOGIC_SAT ⪯ₘ DIO_ELEM_SAT ⪯ₘ DIO_SINGLE_SAT ⪯ₘ H10 ⎭.
> ```
> 
> a student/reader would require to understand

On principle I would agree with you. However, there is some magic here! I suggest you actually run the statement in an IDE and see for yourself what Coq actually displays when it has parsed it ;-)

DmxLarchey:
> I suggest this PR to be merged soon, before further changes.

I am still waiting on @yforster's review. I could merge it myself though.

DmxLarchey:
>     * To be honest (since I'm not too experienced with ltac), I have an intuition, but do not fully understand what is going on in the background just to formulate the reduction chain statement.

The way I understand it is that a chain say `⎩ A ⪯ₘ B ⪯ₘ C ⪯ₘ D ⎭` is first _parsed_ as the tuple `(A,(B,(C,D)))` (because `A ⪯ₘ B` is a right associative notation for `pair A B`) and then _mapped_ to `A ⪯ B /\ B ⪯ C /\ C ⪯ D` with the `ltac` macro.

yforster:
Done!



> a student/reader would require to understand

I think I'm with Dominique here: Yes, the notation uses several hacks. But since it (1) is very suggestive and close to what you'd write on paper and (2) immediately shows the "right" goal when the proof is entered, nobody needs to understand the definition of the notation. 

We could re-implement the notation to only use pure Coq functions, by adding a symbol in front, e.g. `! ⎩ A ⪯ₘ B ⪯ₘ C ⪯ₘ D ⎭`, and basing everything on "true" heterogeneous lists instead of tuples, but unlike the current definition by Maximilian that would not reduce away to the "right" statement immediately, so I'd even prefer to leave it as is.

yforster:
And I forgot to say: Amazing work, thanks Dominique!

> Notice that most of the `Undecidability` library _does not_ follow those guidelines ATM.

You're very right :) I'm happy to see that we're improving though. I'm working on streamlining the `L` and `TM` directories to result in something closer to the guidelines. Especially, I want short files defining the halting problems without any distracting stuff around. Should be done by the end of the week!

DmxLarchey:
> everything on "true" heterogeneous lists instead of tuples, but unlike the current definition by Maximilian ...

Indeed that was roughly my first attempt but Maximilian's solution is _way better_ because now you can directly use the statement to very easily prove either sub-chains or else parts of longer chains. I did put additional simple _proving_ tactics just for that. See e.g. `PCP/Reductions/HALT_TM1_to_PCPb.v`.

And thanks for the thumb up ;-)

PR #56, opened 2020-07-29
BinaryStackMachines to StackMachines

DmxLarchey:
Migration:
+ `BinaryStackMachines` to the _existing_ `StackMachines` directory
+ removal of `omega`, replaced by `lia` in those moves files.

### DIFF
--- a/theories/ILL/ILL_undec.v
+++ b/theories/ILL/ILL_undec.v
@@ -14,9 +14,9 @@ From Undecidability.Shared.Libs.DLW
 
 Require Import Undecidability.Synthetic.Undecidability.
 
-From Undecidability.PCP                  Require Import PCP PCP_undec.
-From Undecidability.BinaryStackMachines  Require Import BSM.
-From Undecidability.MinskyMachines       Require Import MM.
+From Undecidability.PCP              Require Import PCP PCP_undec.
+From Undecidability.StackMachines    Require Import BSM.
+From Undecidability.MinskyMachines   Require Import MM.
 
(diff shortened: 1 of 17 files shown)


yforster:
Looks good!

W.r.t. the `omega` to `lia` replacement: Did you encounter any issues? Also on my TODO list for this week is to port everything to 8.12, that way we can release an (internal) opam package once the MetaCoq package is ready. I'm still hoping that some `grep` and `sed` magic will make the `omega` to `lia` replacement easy, but I don't know how often incompatibilities occur.

DmxLarchey:
> some `grep` and `sed` magic will make the `omega` to `lia` replacement easy, but I don't know how often incompatibilities occur

I can tell you that `lia` as a drop-in replacement for `omega` works most, _nearly_ all the time. But once I made the test change `Ltac omega := lia` on the `MM` part of the code base, and then I did encounter one or two occurrences where this failed. At the time, I was not in the mood of investigating so I postponed. It was in one of the compilers, I think `µ-rec` to `MM(nat)`. 

I guess you could try something like

```
Ltac old_omega := omega.
Ltac omega := (lia || (idtac "lia failed !!"; old_omega)).
```
and see which files have problems ?

DmxLarchey:
Travis says OK. I merge.

DmxLarchey:
Now that I am updating more  `omega` to `lia`, I realize that I often do `try omega` and then `try lia` might produce strange side effects if it does not solve exactly the same subgoals as `try omega`.

PR #57, opened 2020-07-29
Replace Omega by Lia (in many places but not everywhere ...)

DmxLarchey:
Let us open it that way: `fgrep Omega $(find -name "*.v")` and `fgrep omega $(find -name "*.v")` are your new friends!

`Omega`/`omega` have been completely removed in all the `.v` files of following directories
+ `StackMachines`
+ `MinskyMachines`
+ `FRACTRAN`
+ `ILL`
+ `MuRec` (this dir needs refactoring **badly**)
+ `H10`
+ `Shared/Libs/DLW`

Notice `omega` might still appear in commented out code eg  `(* ... omega ... *)`.

Some observations:
+ sometimes `lia` solves more goals than `omega`, never observed the converse.
+ `ZArith` seems to silently activate `omega` so a negative `greg Omega` does not mean `omega` is not used
+ `Undecidability.Shared.Prelim` seems to export `omega`.
+ `lia` might solve goals using **less** hypotheses, changing (strengthening) the statement if those hypotheses where declared inside a `Section`.

**Main lesson learned**:
+ `lia` **always** solved the goals that `omega` solved in the above directories, and e.g. in `Libs/DLW` there were 1000s fu..ing of them.
+ sometimes, `lia` solves **more goals** than `omega` which _could be problematic_ with script like `...; try lia; ...` because there are now fewer sub-goals after `lia` than there were with `omega`, meaning that the proof script now contains sub-proofs for sub-goals that do not exist any more.

I do hope this PR motivates @yforster from the remaining parts of the library ;-)


yforster:
We really, really have to get rid of `Undecidability.Shared.Prelim`...

I'll merge this PR and feel motivated to work on omega -> lia :) Probably I'll do it in one go with porting to 8.12, but first I finish the nice defs for L and TMs.

DmxLarchey:
Good luck Yannick!! B.t.w., compile time on `CI` did not change significantly by replacing `omega` with `lia`.

DmxLarchey:
And I will look into the `Prelim` usage in the directories that I master. I think it is mainly use for notations.

PR #59, opened 2020-07-30
Reorganising TM

yforster:
I started reorganising TM.

There is a nice `TM/TM.v` file now (but it still needs to be commented). I moved all the cluttering files in `TM` to `TM/Util` and removed stuff in `Reductions`. There are some leftovers in `Problems` I think, I'll do that tomorrow. I'm not so sure what to do about the directories that belong to the framework, e.g. `TM/Core` or `TM/Univ` on the long run.

I still want to add comments and try to port away from omega before we merge this PR, but it can't hurt if you already have a look @DmxLarchey @mrhaandi.

Next up then `L`, which will change things in `TM` as well because I'd like to get rid of `LAM`.


DmxLarchey:
I am commenting/questionning the code directly. I won't make changes except for comments.

DmxLarchey:
I think there is an issue with empty cells in between written cells on tapes but maybe I am missing something here.

mrhaandi:
This is so much cleaner now than the previous definition, good work.
Regarding presentation:
you use for function names `mv`, `wr`, `current` and for variable names `tp`, `ls`, etc. so there is no scheme to see whether an object is a local variable (usually short names) or a function (usually longer names), cf. `mv m (wr c tp)) t a`.
Regarding conformity with literature:
- a common construction is from a starting state to overwrite the input with a fixed input (e.g. the empty input, i.e. `niltape`). This looks not possible with the current definition.
- usually, if the transition function declares a move, e.g. `R`, then the step relation has to obey it.
This is not true for the current definition in the `rightof` position.
- some constructions actively use the blank symbol to store information (truly two-symbol machines, i.e. {blank, `0`}).
This is not possible here.

I fully understand the design choices behind the above points and it is ok to leave as is, but it has to be noted, e.g. in a comment.

DmxLarchey:
Of course I agree that this is a much cleaner presentation than what existed before. Thanks @yforster. 

But I think that the definition of `TM` is a bit special. Since we somehow use it as the _ground seed_ for the synthetic undecidability definition, the definition of `HaltTM1` (and hence of `TM`) has to be unambiguously understood as legitimate. I think there are issues here: 
+ eg, the definition of [_Hopcroft & Ullman_](https://en.wikipedia.org/wiki/Turing_machine) as described on Wikipedia allows for blank symbols to occur anywhere, not only on the left/right edge of regular symbols.
+ I understand that `TM` are tricky to formalize because they are not so precisely characterized in the literature. For instance, in the above definition, the transition function is a _partial function_. It is not said to be a _table_. Clearly one cannot allow the transition function not to be computable.

yforster:
Please no more commits on this branch, otherwise I get into merging hell :) I'll answer to your questions in a moment.

yforster:
I'm not done, but we can get the discussion started:

> (* DLW: why not rightof :  list Σ -> Σ -> tape ? *)

That would also have been an option. But the whole CPP 20 paper was built around this definition, I think because Asperti & Ricciotti picked it like this.

In general I'm open to changing the definition of Turing machines to the outside, and have an equivalent internal version for the framework. But that only makes sense if we find a simpler but equivalent definition of Turing machines which is significantly different from the current one.

> (** DLW: Are you sure you want one letter constructors *)

I left them as is because I thought it would be hard to change them. Some `git grep` and `sed` helped however, they are renamed into `Lmove | Nmove | Rmove` :+1: 

>       a tape like ..... _ _ a b _ c d _ _ f e _ _ ...
>      would not be allowed
>      Not sure this corresponds to usual literature?

It does not correspond to the *usual* literature indeed. It corresponds to the CPP '20 paper though :) I inserted a sentence above, explaining that there are no blank symbols in the representation and that blanks have to be explicitly part of the alphabet if one wants to work with them.

> (* DLW: I suggest state instead of state, trans and start feels strange with state. *)

Indeed, thanks.

>  (* DLW: is there a reason for not always using tapes Σ n below ? *)

That was a typo, it should never be tapes.


> I understand that `TM` are tricky to formalize because they are not so precisely characterized in the literature. For instance, in the above definition, the transition function is a _partial function_. It is not said to be a _table_. Clearly one cannot allow the transition function not to be computable.

Computability is irrelevant for the transition function! It's a function with finite domain and finite co-domain, such functions are always computable, even constructively.

DmxLarchey:
> Computability is irrelevant for the transition function! It's a function with finite domain and finite co-domain, such functions are always computable, even constructively.

Well it depends on what you call _function_ but for sure there are non-computable functional relations `Rf : X -> Y -> Prop` where `X` and `Y` are `finType`. There are even non-computable relations `unit -> unit -> Prop`.

DmxLarchey:
> there are no blank symbols in the representation

So there is a distinction between _blank_ and _empty_ which is not always the case in the literature.

I have problems with your definition of `mv` because moving `n` times left and then `n` times right is not the identity, even extensionaly.

yforster:
> > Computability is irrelevant for the transition function! It's a function with finite domain and finite co-domain, such functions are always computable, even constructively.
> 
> Well it depends on what you call _function_ but for sure there are non-computable functional relations `Rf : X -> Y -> Prop` where `X` and `Y` are `finType`. There are even non-computable relations `unit -> unit -> Prop`.

Okay, for relations you don't get far constructively indeed. My intuition is however that classically, finite *total* functional relations are computable. If you're allowed to be partial, you're right, there are non-computable relations.

yforster:
> > there are no blank symbols in the representation
> 
> So there is a distinction between _blank_ and _empty_ which is not always the case in the literature.
> 
> I have problems with your definition of `mv` because moving `n` times left and then `n` times right is not the identity, even extensionaly.

Very good point, that has to be made clear in the comments. What is the identity is: "do the following n times: write the current symbol of it exists, blank if it doesn't exist, then move to the right" and then "move n symbols to the left". So the behavior of a normal move to the right can be simulated once you have a blank symbol, but our representation is more fine grained.

mrhaandi:
> It is, you just need the blank in the alphabet. I know I'm repeating myself, but a blank can occur anywhere once there are blanks.

The difference is that blank as part of the alphabet behaves differently compared to `None` as blank as part of an untouched tape (again, right move on `rightof` is the culprit). So having a blank as tape symbol introduces a second, different blank notion.
Of course, one can simulate `None`-blanks by symbol blanks. However, I do not see how one can have a truly two-symbol machine, as the `None`-blanks can be used to store some information.
Also, non of this matters right now, but I consider some minimal constructions in the future.

yforster:
> However, I do not see how one can have a truly two-symbol machine, as the `None`-blanks can be used to store some information.

A "true" two-symbol machine would correspond to a one-symbol machine in traditional presentations? I.e. one symbol plus a blank? That can indeed not be simulated in our setting.

DmxLarchey:
I understand that this is the CPP'20 paper cleaned up which is good. But you must admit that in general we cannot pick up one definition (in this case `TM`) and say this is our definition. What would be the guarantee that we do not alter the definition in that case? We have to convince that the definition conforms the well-established definitions. This is a tricky job because one description is _informal_ and the other one is _formal_, so this will never be fully mechanisable.

I have no-problem with using alternate representations because they are more convenient to work with. It is both necessary and relevant for obtaining manageable proofs.

So we can perfectly keep this cleaned up definition of `TM` corresponding to CPP'20 but I think we need to devise at least an alternate def. that can be compared to at least one _autoritative definition_ in the computability literature and build a bridge (bi-reductions) between those two definitions. Self-reference cannot really be admitted as _autoritative_.

I even think one memo could be written collecting the reasons why it is difficult to implement Turing machines in constructive type theory.

DmxLarchey:
> Very good point, that has to be made clear in the comments.

Well this is my main issue with this definition. It clearly goes _against_ my intuition.

DmxLarchey:
Another related but _less important_ issue is whether equivalence of tapes should be strongly decidable? By equivalence I mean `t ~ t'` iff when you perform the same sequence operations `mv`/`wr` on `t` and `t'` and then a `rd`, you always read the same value. 

With the CPP'20 definition, I have the impression equivalence is identity. But that need not be the case with any representation, typically a map `t : Z -> Σ` (`Z` are relative integers). 

If you just assume the domain where `t` is not void to be bounded by some/`exists` `n`, then `equivalence` is logically decidable but not computationally decidable. With the CPP'20 `TM` definition, the bound is hard-coded and can be computed from the representation of tapes. So equivalence/identity is strongly decidable in this case.

Not even speaking whether tapes can contain an infinite mount of information.

DmxLarchey:
> Please no more commits on this branch, otherwise I get into merging hell :) I'll answer to your questions in a moment.

Ok ok. You can remove the `tape.v` file if you want. It is not listed as part of `_CoqProject`. Just a set of ideas about _abstract_ tapes and what kind of axioms/properties they should satisfy.

mrhaandi:
> But that need not be the case with any representation, typically a map `t : Z -> Σ` (`Z` are relative integers).

There are papers clearing up that the tape representation `t : Z -> Σ` is part of _generalized instantaneous description_ compared to _instantaneous description_ that has only finitely many non-blank symbols. So, `t : Z -> Σ` is not traditional.

yforster:
I think in general we're all on the same page.

Final comment:

> I have problems with your definition of `mv` because moving `n` times left and then `n` times right is not the identity, even extensionaly.

That's not a thing TMs as defined in Hopcroft & Ullman can do! They always have to write something before moving. So yes, the culprit is the action of `Rmove` on `rightof`, but only because it is possible to end up in `rightof` after `wr`. If one would change the definition of `wr`, the problem goes away.

I tried thinking about simpler definitions. The main issue with all nice definitions is, that proving the halting problem *equivalent* to the present one might be not straightforward. One direction is always easy. Given the historical fact that this is the version of TMs we published in a peer-reviewed paper I'm going with the solution to explain the difference to traditional presentation, but without making a traditional definition precise and showing the equivalence.

I suspect that when reducing TM halting via a compilation to single-tape machines with boolean alphabet to BSMs I'll need an intermediate step with a less tedious tape representation which you can reduce *to* from TMs, but the other direction will be hard. Since it's an internal step in the proof, I won't make that alternative definition prominent I think. I hope to get to that next week.

yforster:
I commented the TM.v file and renamed L, R, and N. I still want to rename mTM into TM, then this PR might be good to go.

DmxLarchey:
What about the following definition of tapes

```
  (** tapes with blank cells: 
       - a,b,c for regular symbols in Σ
       - ?,x for either blank or reg. in option Σ
       - _ for blank in option Σ *)

  (* it_one x represents 
       a) either empty 
             _ _ [_] _ _    as  (it_one None)
       b) or exactly one non blank under the head
             _ _ [a] _ _    as (it_one (Some a))

     it_lft represents tapes which are entirely left of the head
          _ _ a ? ? ? [x] _ _   as (it_lft a [?;?;?] x)

     it_rt represents tapes which are entirely right of the head
          _ _ [x] ? ? ? b _ _   as (it_rt x [?;?;?] b)

     it_both represents tapes which are both left and right of the head
          _ a ? ? [x] ? ? b _   as (it_both a l x r b)

   *) 

  Inductive itape : Type :=
    | it_one  : option Σ -> itape
    | it_lft  : Σ -> list (option Σ) -> option Σ -> itape
    | it_rt   : option Σ -> list (option Σ) -> Σ -> itape
    | it_both : Σ -> list (option Σ) -> option Σ -> list (option Σ) -> Σ -> itape.

  (* Moving left *)

  Definition mv_lft (t : itape) :=
    match t with
      | it_one None                     => it_one None                     (*  _ _ [_] _ _ ~~> _ [_] _ _ _ *)
      | it_one (Some a)                 => it_rt None nil a                (*  _ _ [a] _ _ ~~> _ [_] a _ _ *)

      | it_lft a nil         None       => it_one (Some a)                 (*  _ a [_] _ _ ~~> _ [a] _ _ _ *)
      | it_lft a nil         (Some b)   => it_rt (Some a) nil b            (*  _ a [b] _ _ ~~> _ [a] b _ _ *)
      | it_lft a (None::l)   None       => it_lft a l None                 (*  a _ [_] _ _ ~~> a [_] _ _ _ *)
      | it_lft a (x::l)   (Some b)      => it_both a l x nil b             (*  a x [b] _ _ ~~> a [x] b _ _ *)
      | it_lft a (Some b::l) None       => it_lft a l (Some b)             (*  a b [_] _ _ ~~> a [b] _ _ _ *)

      | it_rt x l a                     => it_rt None (x::l) a             (*  _ _ [x] ? a ~~> _ [_] x ? a *)

      | it_both a nil x r b             => it_rt (Some a) (x::r) b         (*  _ a [x] ? b ~~> _ [a] x ? b *)
      | it_both a (x::l) y r b          => it_both a l x (y::r) b          (*  a x [y] ? b ~~> a [x] y ? b *)
    end.

  (* Moving right *)

  Definition mv_rt (t : itape) :=
    match t with
      | it_one None                     => it_one None                     (*  _ _ [_] _ _ ~~> _ _ _ [_] _ *)
      | it_one (Some a)                 => it_lft a nil None               (*  _ _ [a] _ _ ~~> _ _ a [_] _ *)

      | it_lft a l x                    => it_lft a (x::l) None            (*  a ? [x] _ _ ~~> a ? x [_] _ *)

      | it_rt None     nil     a        => it_one (Some a)                 (*  _ _ [_] a _ ~~> _ _ _ [a] _ *)
      | it_rt (Some a) nil     b        => it_lft a nil (Some b)           (*  _ _ [a] b _ ~~> _ _ a [b] _ *)
      | it_rt None     (None::l) b      => it_rt None l b                  (*  _ _ [_] _ b ~~> _ _ _ [_] b *)
      | it_rt (Some a) (x::l) b         => it_both a nil x l b             (*  _ _ [a] x b ~~> _ _ a [x] b *)
      | it_rt None     (Some a::l) b    => it_rt (Some a) l b              (*  _ _ [_] a b ~~> _ _ _ [a] b *)
 
      | it_both a l x nil b             => it_lft a (x::l) (Some b)        (*  a ? [x] b _ ~~> a ? x [b] _ *)
      | it_both a l x (y::r) b          => it_both a (x::l) y r b          (*  a ? [x] y b ~~> a ? x [y] b *)
    end.

  (* Left/right moves are strict inverse to each other *)

  Fact mv_lft_rt t : mv_lft (mv_rt t) = t.
  Proof. now destruct t as [ [] | ? [ | [] ] [] | [] [ | [] ] ? | ? [] [] [] ? ]. Qed.

  Fact mv_rt_lft t : mv_rt (mv_lft t) = t.
  Proof. now destruct t as [ [] | ? [ | [] ] [] | [] [ | [] ] ? | ? [] [] [] ? ]. Qed.

  (* I think one can also show that equivalent tapes are identical with that definition *)
```

DmxLarchey:
I leave this here ...

```
  Variable x : Σ.

  Let t := leftof x (x::x::nil).

  Eval compute in mv Rmove ((mv Lmove t)).
  Eval compute in mv Rmove ((mv Rmove (mv Lmove ((mv Lmove t))))).
  ```

**Comment by mrhaandi** [comment]
> What about the following definition of tapes

This has the drawback going from a 7-case move definition to a 20-case one.
By the way, traditional machines with tapes only infinite to the right do not have inverse left-right moves, because one cannot move left from the starting position.

**Comment by mrhaandi** [comment]
If I may put my two cents in, what about a simplified tape representation (resembling the current one) as `(list Σ) * (option Σ) * (list Σ)`.
Specifically, 
- `(nil, None, nil)` is `niltape `
- `(a :: l, None, nil)` is `rightof a l` (similarly `leftof`)
- `(l, Some a, r)` is `midtape l a r`
Then, the definition of `write a` is just `(l, _, r)` -> `(l, Some a, r)` possibly eliminating a case distinctions in the proofs.
Of course, this would be just a reformulation of current code and not change the blank vs. `None` situation.

**Comment by yforster** [comment]
> If I may put my two cents in, what about a simplified tape representation (resembling the current one) as `(list Σ) * (option Σ) * (list Σ)`.
> Specifically,
> 
>     * `(nil, None, nil)` is `niltape `
> 
>     * `(a :: l, None, nil)` is `rightof a l` (similarly `leftof`)
> 
>     * `(l, Some a, r)` is `midtape l a r`
>       Then, the definition of `write a` is just `(l, _, r)` -> `(l, Some a, r)` possibly eliminating a case distinctions in the proofs.
>       Of course, this would be just a reformulation of current code and not change the blank vs. `None` situation.

I had this in paper a couple of days ago.

My issue was that `([a], None, [a])` is very weird. You can easily reduce the Italian tapes (they are due to Asperti and Ricciotti after all) to these triple representation, but not vice versa - unless you're willing to define valid tapes and exclude the case where the head points to empty but the rest is written.

**Comment by mrhaandi** [comment]
> My issue was that `([a], None, [a])` is very weird.

Agreed.

**Comment by yforster** [comment]
Observing the issue and fixing it in the most straightforward way, one ends up with either `{ '(l, m, r) : list Σ * option Σ * list Σ | match l, m, r with _ :: _, None, _ :: _ => False | _, _, _ => True end}`, but then has to inline proofs in the definition of moves, or  `{o : option Σ & match o with Some => list Σ * list Σ | None => unit + Σ * list Σ + Σ * list Σ end}`, which is more or less the Italian representation, just uglier.

**Comment by yforster** [comment]
I got confused and pushed the branch without `omega` and a concise definition of L-halting here.

Removing `omega` was no major pain. There was just one proof in third-order unification which was a bit harder to close, apart from that there were ~10 or so occasions where `lia` closed more than `omega`.

I'd like to merge this sooner rather than later, even if we want to discuss some details around `TM` and `L`.

**Comment by DmxLarchey** [comment]
> I'd like to merge this sooner rather than later, even if we want to discuss some details around `TM` and `L`

Yes of course. But I think we are not done with Turing Machines...

**Comment by DmxLarchey** [comment]
> classically, finite _total_ functional relations are computable

`AC : Prop` is eg _axiom of choice_ (or any other logical axiom)

```
R (_ : unit) (y : bool) := y = true /\ AC \/ y = false /\ ~ AC
```

is a finite functional relation, moreover _classically_ total, and not it is not computable.

yforster:
Classically, it is computable: Either AC holds, then the constant true function does the job. Or AC doesn't hold, then take the constant false function.

DmxLarchey:
What I mean of course is: it is not Coq-computable ;-).

Using finite choice (over listable & discrete source `X` type), you can prove that there `exists` a reifer, ie you can build a proof of `exists f : X -> Y, forall x, R x (f x)`. You only need `XM` to show that the above `R : unit -> bool` is _total_.

But you cannot Coq-compute such an `f`, i.e. you cannot build a term `{ f : X -> Y | forall x, R x (f x) }`, even assuming `XM`.

But well I guess you know that already.

PR #60, opened 2020-08-04
Branch for Coq 8.12

yforster:
There's a `coq-8.12` branch now which I can compile locally, works on Travis and compiles on Andrej's Windows machine. I didn't use PRs to create it because there was nothing to merge against, so I'm opening this PR to have a central point for discussion.

The opam packages for `smpl` and the `base-library` are done. The official MetaCoq opam package is on its way, I created an internal version to make it easier to install. It's important to run `opam update` first, afterwards `opam install . --deps-only` should install the necessary packages.

There are some more warnings now, but not too many, I'll try to fix them in later PRs.

I'd also like to brush up the README a bit, feel free to mention relevant feedback here. I added the recent papers and fixed some links in the problems list.

We could add a `Troubleshooting` section and include hints how to use the library under Windows and how to use `vscoq` and WSL together. @mrhaandi do you have suggestions for concrete bits of text to put there?

If there are no requests I'd also like to make `coq-8.12` the default branch.

@fakusb, it might be easiest to port `coq-library-complexity` quickly so we can fix things in our library that you need. The representation of constant names in MetaCoq changed significantly (paths are now explicitly represented), which means I had to touch various parts of the L extraction. Also `lia` behaves differently on 8.12. If you encounter bugs and report them there are good chances the fixes will make it into 8.12.1. Andrej encountered two lia regressions in his code which we're currently workarounding.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,13 +7,13 @@ Feel free to contribute or start using the problems!
 
 ## Existing undecidable problems
 
-- Post correspondence problem (`PCP` in [`Problems/PCP.v`](theories/Problems/PCP.v)), **`good seed`**
-- Halting problems for single-tape and multi-tape Turing machines (`Halt` in [`Problems/TM.v`](theories/Problems/TM.v))
+- Post correspondence problem (`PCP` in [`PCP/PCP.v`](theories/PCP/PCP.v)), **`good seed`**
+- Halting problems for single-tape and multi-tape Turing machines (`HaltTM` in [`TM/TM.v`](theories/TM/TM.v))
 - Halting problem for Minsky machines (`MM_HALTING` in [`Problems/MM.v`](theories/Problems/MM.v))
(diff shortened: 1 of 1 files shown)


mrhaandi:
> @mrhaandi do you have suggestions for concrete bits of text to put there?

Something like:
If you use Visual Studio Code on Windows 10 with Windows Subsystem for Linux (WSL), then _local_ opam switches may cause issues.
To avoid this, you can use a non-local opam switch, i.e. `opam switch create 4.09.1+flambda`.

mrhaandi:
In `Troubleshoot` you should also say something along the lines
- if a previous Coq version is used, then the particular corresponding branch needs to be checked out
- due to compatibility issues, not every branch contains exactly the same problems

By the way, there are currently a lot of confusing branches such as `my_own_tapes` etc. in the main repository. I'd suggest to keep the branch structure in the main repository clean (maybe only Coq-x.xx and noopam for extreme cases) and use forks and PRs for development.

PR #61, opened 2020-08-12
Binary Turing machines and direct reduction from Turing machines to BSM

yforster:
I have an alphabet reduction from arbitrary alphabet to binary alphabet (reasonably easy, very nice application of the TM framework) and a simulation of single-tape binary Turing machines (`SBTM`) on binary stack machines (reasonably easy, very nice application of the `sss` framework). @DmxLarchey:The simulation uses 4 stacks: Left part of the tape, middle part of the tape, right part of the tape, and empty register for unconditional jumps. One can optimise middle away by merging it into the state space, but I didn't bother to, mainly because the compiler to MMs adds two counters anyways.

@mwuttke97 @fakusb The only thing that was slightly annoying me regarding TMs is that we do not have support to prove TMs total without actually talking about step indices. While one can abstract away from labels regarding realisation, abstracting away from steps regarding termination is impossible. It was no big deal, as an effect of this I also have a proof that the simulation is linear. If we wanted (I don't want to) we could do a similar analysis for the SBTM -> BSM compilation. @fakusb  that might be useful for you in case you want to bring TMs to register machines: You can start at BSM, which have a nicer structure for compilation.

Overall 2.5 days for the first part and 2.5 days for the second part, so quite a good investment :)

Two TODOs left: Change the `BSM_undec` proof and add an `SBTM_undec` file.

@mrhaandi @DmxLarchey SBTMs can be taken as alternative definition for presentation regarding Turing machines. They are less involved than TMs because they (1) do not mention vectors, (2) work with `Fin.t` instead of `finType` and (3) work with `list bool * option bool * list bool` as tapes. The issue still is that `([true], None, [true])` is considered a tape, but that's strange. Since `SBTM` is just an intermediate step to `BSM`, that's fine for me.

### DIFF
--- a/theories/Shared/FinTypeEquiv.v
+++ b/theories/Shared/FinTypeEquiv.v
@@ -1,0 +1,80 @@
+From PslBase Require Import FiniteTypes.
+
+Fixpoint position {X : eqType} (x : X) (l : list X) : option (Fin.t (length l)).
+Proof.
+  induction l.
+  - exact None.
+  - cbn. decide (a = x).
+    + exact (Some Fin.F1).
+    + destruct (position _ x l) as [res | ].
+      * exact (Some (Fin.FS res)). 
+      * exact None.
+Defined.
+
+Lemma position_in {X : eqType} (x : X) (l : list X) (H : x el l) : 
+  { i | position x l = Some i}.
+Proof.  
+    induction l; cbn in *.
+    - inv H.
+    - decide (a = x).
+        + eauto.
+        + destruct IHl as [i IH]. firstorder. rewrite IH. eauto.
+Defined. 
+
+Definition posIn {X : eqType} (x : X) (l : list X) (H : x el l) : Fin.t (length l).
+Proof.
+    eapply position_in in H. destruct (position x l) as [i | ]. exact i.
+    abstract (exfalso; firstorder congruence).
+Defined.
+
+Fixpoint getat {X : Type} (l : list X) (i : Fin.t (length l)) : X.
+Proof.
+    destruct l.
+    - inv i.
+    - cbn in i. eapply (Fin.caseS' i (fun _ => X)).
+      + exact x.
+      + eapply getat.
+Defined.
+
+Arguments getat {_} _ _.
+
+Lemma getatIn {X : Type} (l : list X) (i : Fin.t (length l)) : 
+    getat l i el l.
+Proof.
+    induction l.
+    - inv i.
+    - cbn in *. eapply (Fin.caseS' i); cbn. eauto. eauto.
+Qed.
+
+Lemma finite_n (F : finType) :
+    { n & {f : F -> Fin.t n & { g : Fin.t n -> F | (forall i, f (g i) = i) /\ forall x, g (f x) = x }}}.
+Proof.
+    destruct F as (X & [l H]). cbn in *. 
+    exists (length l).
+    assert (Hin : forall x, x el l). { intros x. eapply count_in_equiv. rewrite H. lia. }
+    exists (fun x => proj1_sig (position_in (Hin x))). exists (@getat _ l). split.
+    - intros i. destruct position_in. cbn.
+      specialize (H (getat l i)). clear - H e. 
+      induction l.
+      + inv x.
+      + cbn in *. revert H e.
+        eapply (Fin.caseS' i). cbn.
+        * decide (a = a); congruence.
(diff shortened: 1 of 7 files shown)


DmxLarchey:
Nice work. Feedback on the `sss` framework is very welcome !!

yforster:
I had an issue with proving `<sc` automatically and added an External hint locally in the reduction file. If you want you can have a look and tell me whether there is a nicer way.

I would have appreciated tactics of the sort `bsm sss auto step`, because oftentimes it is inferrable what's the next instruction to run, in particular in `MOVE_L` and `MOVE_R`. Similarly but simpler, it would be nice if one does not have to specify the labels when using `bsm sss POP`, i.e. if one can just write `bsm sss POP 0`, or even `bsm sss POP`.

Apart from that I got a bit lost what is what amongst `sss_terminates`, `sss_progress`, `sss_computes` and how they are interconnected, but it's easy enough to find out what's going on using Print and Search. It helped a lot to look at your actual `PCP` to `BSM` reduction there to find the right structure.

DmxLarchey:
Yes concerning the automatic computation of steps of programs, the problem I faced with to find which instruction is executed. Ideally, one who find it using the `PC` value but the problem is that programs often looks like `L1++i1::i2::L2++i3::L3` where the length of `L1`, `L2`or `L3` may be parametrized by `nat` parameters, ie you could have `length L1 = 2*u+v+30` ... 

So at the time I designed the framework, I found it easier to identify say `i2` above using its parameters, provided they are syntactically different from those of `i1` and `i3`. This is often the case but not always so
I even had to make them look different (with `Let` defs) just that they get properly identified.

Then the idea of the framework is first to rewrite the program as `L1++[i1]++[i2]++L2++[i3]++L3` and then to focus on say `i2` by rewriting to `(L1++[i1])++[i2]++(L2++[i3]++L3)`. Then you can use Coq unification mechanism to correctly execute `i2` by matching of the pattern `_++[_]++_` ...

An alternate approach would of course be to use the `PC` value to find `i2` by choosing the correct split between 

+ `L1++[i1]++([i2]++L2++[i3]++L3)`
+ `(L1++[i1])++[i2]++(L2++[i3]++L3)`
+ `L1++[i1]++[i2]++L2)++[i3]++L3`

so that the `PC` constraint is solvable by `lia`. A bit of `Ltac` programming that I thought was not worth investing in before because of limited usage. But may be my first impression was not correct... 

In principle Coq should obviously be able to execute by itself each step of those little programs with minimal user input, may be just the number of steps to execute.

DmxLarchey:
And thank you indeed for the nice feedback !!

mwuttke97:
> The only thing that was slightly annoying me regarding TMs is that we do not have support to prove TMs total without actually talking about step indices. While one can abstract away from labels regarding realisation, abstracting away from steps regarding termination is impossible

Do you mean that you want to just prove termination of a machine without specifying the number of steps? This is not impossible, but I've never implemented it. It can be done by introducing a new '1-ary' relation `terminates : tapes n -> Prop`. One can prove the same theorems for `TerminatesIn` for this new definition. (The 'relation' for `While` will be co-inductive again, etc.)

Btw, I just saw that you moved the definitions from `TM.v` -- nice :-) The comment in line 3 `TM/Util/TM_facts.v` should be removed though; there is hardly anything significant from A&R in this file and the attribution line is already included in `TM/TM.v`.

yforster:
> Do you mean that you want to just prove termination of a machine without specifying the number of steps?

@mwuttke97 Yes, exactly. I am aware that one can define the notion (A&R do it I think) and was hoping that one can get the corresponding lemmas for our building blocks for free from our stronger lemmas, but that seems to be not the case.

Because I wasn't interested in re-stating the lemmas for the framework I resorted to semi-ugly existential quantification and evar manipulation. It works, but the code might not be very stable in future versions of Coq.


@DmxLarchey I introduced an `eval` relation fo Turing machines and defined the halting problem for Turing machines with it, but everything else regarding Turing machines is still defined with `loopM`. I am happy with this situation, because I like that we have a simply, compact, and self-contained definition exposed to the user, and a more flexible one used internally.

Similarly, I'd like to define an evaluation relation on `BSM`s and `MM`s, define the halting problem using this relation and them immediately show it equivalent to the ss semantics. The benefit is that one does not have to check the sss files to understand the definition of `BSM` and `MM` halting. The only things in the code that change are points where the halting problems are explicitly mentioned, there we then have to change to the alternative halting problem based on sss to go on. What do you think?

PR #64, opened 2020-10-02
Backporting improvements in L and TM-framework

fakusb:
This ports improvements for L and TM framework to merge back into undecidability repository.

This PR is work in progress and gets tidied up more. I want to have a draft here so CI runs.


theories/_CoqProject
```diff
@@ -331,20 +331,19 @@ TM/Univ/StepTM.v
...
-# TM/PrettyBounds/UnivBounds.v
-# TM/PrettyBounds/M2MBounds.v
+TM/PrettyBounds/BaseCode.v
+TM/PrettyBounds/UnivBounds.v
+TM/PrettyBounds/M2MBounds.v
 
-# TM/PrettyBounds/SpaceBounds.v
-# TM/PrettyBounds/BaseCodeSpace.v
-# TM/PrettyBounds/UnivSpaceBounds.v
+TM/PrettyBounds/SpaceBounds.v
```
yforster, line 338: Are these files really needed for compilation? On the long run we shouldn't have such proofs in the undecidability library. Even on short term it might make sense to already now move them to the complexity library.

fakusb: Yes, we should move them out.

fakusb:
So, I removed all files concerning "nice bounds" on TMs.

yforster:
Thanks! Once CI says yes I'll merge

PR #65, opened 2020-10-03
TM framework: speedup (?)

fakusb:
This draft is for some speedups in the TM framework. I want to have CI time this for comparison.

Do not merge before #64 . Even after, probably rebased ontop of #64.


fakusb:
This and local tests seem to imply that I get 2 minutes  speedup for the TM-related proofs.
The worst offender, the TM interpreting L, goes from 3:21min to 2min.

yforster:
I'll wait with reviewing this code until we have finished merging the other `PR`. Alternatively you could change the merge target to your other PR, that way we get a nicer diff in the `Files changed` overview

fakusb:
I rebased upon the current state of the branch. We still edit 28, but only because the internal names in the TM-framework-using proofs changed and I had to adjust them. This is due to me changing the used tactics.

PR #68, opened 2020-10-07
Reductions for undecidability of fragments of Classical Linear Logic

DmxLarchey:
Extensions of the results obtained for Intuitionistic Linear Logic `ILL` (see _CPP'19 paper_) to Classical Linear Logic `CLL`. We add undecidability for two fragments of _cut-free `CLL`_:

+ the fragment limited to the `{!,&,⊸}` connectives, also denoted `rCLL` in here;
+ the full `CLL`.

Both are obtained by embedding `EILL` into them, possibly through `rILL` and `ILL` respectively. We use H. Schellinx  observation (_JLC 91_) that `CLL` is conservative over any syntactic fragment of `ILL` that does not contain both `⟘` and ` ⊸`, which is the case of `EILL`. 

Every syntactic fragment in between `rCLL` and `CLL` is also undecidable but we need to implement cut-elimination for this and this is not done at the moment.


mrhaandi:
This PR adds a proxy file `CLL.v` to the top-level `Problems` folder, which is discouraged in favor of a well-commented `theories/ILL/CLL.v`.
It is planned to entirely remove the top-level `Problems` and `Reductions` folders in the future.

yforster (approved):

yforster:
I would say we merge this to make it easier for Andrej to prepare his upcoming PR, but if you have a spare hour the next time @DmxLarchey it would indeed be great if we could get rid of the `Problems` and `Reduction` directories. Andrej will take care of many files in his PR, and maybe we can split the rest (first-order and linear logic) between the two of us.

PR #69, opened 2020-10-22
Getting rid of legacy Problems directory

DmxLarchey:
I did remove the following `*.v` in `theories/Problems/`

- `MM.v`
- `MM2.v`
- `BSM.v`
- `ILL.v`
- `CLL.v`
- `FRACTRAN.v`
- `DIOPHANTINE.v`

both the files themselves and the pointers to them in `_CoqProject` and `README.md`.

Remaining is `FOL.v` and `FOLFS.v` and also `RA_UNIV.v`.


mrhaandi:
Sorry for the redundant work, this will be part of a larger cleanup that I'm doing for a current paper (PR to be submitted tomorrow).
I also tried, while removing the files to preserve some comments (putting the comments into the correct file) regarding related literature.

mrhaandi:
Currently only two FOL related files are in each `Problems` and `Reductions` (and can be dealt with in near future), so this particular PR is not needed.

PR #74, opened 2020-10-28
Undecidability for IMSELL with 3 modalities

DmxLarchey:
Two problems:

- `ndMM2`: acceptance for non-deterministic Minsky machines with two counters;
- `IMSELL`: entailment in Intuitionistic Multiplicative Sub-Exponential Linear Logic (see [K. Chaudhuri's paper](https://doi.org/10.1017/S0960129516000293) for the "classical" version of the logic).

`IMSELL` is interesting because it is a step towards `(I)MELL` of which the decidability status is still **unknown and a very difficult** problem. Indeed, it is trivial to encode Petri nets reachability in `(I)MELL` and that decidability proof is _very involved_.

Contrary to `EILL`, `IMSELL` does not include additive connectives but the reduction implements _zero tests_ using three modalities, one of which being universal (i.e. it has weakening and contraction rules). 

Two many-one reductions:

- `ndMM2` is reduced from termination of `MM2` on the zero state (itself already reduced from `FRACTRAN`);
- `IMSELL` is reduced from acceptance for ndMM2 by combining the [encoding of K. Chaudhuri](https://doi.org/10.1017/S0960129516000293) with a _trivial phase semantics for `IMSELL`_ inspired from that of `EILL` in the [CPP'19 paper](https://doi.org/10.1145/3293880.3294096).

And thus two new undecidability results in:

- `CounterMachines/ndMM2_undec.v`;
- `ILL/IMSELL_undec.v`. 

_Update:_ following @mrhaandi suggestion, the statements of the main reduction for `IMSELL` integrate the condition on the signature by packing it with the signature into a Σ-type.


mrhaandi:
@DmxLarchey Is there a reason you put ndMM2 (that are internally documented as Minsky machines) into CounterMachines instead of MinskyMachines?

DmxLarchey:
> @DmxLarchey Is there a reason you put ndMM2 (that are internally documented as Minsky machines) into CounterMachines instead of MinskyMachines?

Yes indeed @mrhaandi . The reason is that `ndMM2` is standalone and does not depend on the `MM` framework at all. Also, the predicate is _acceptance_ and is not defined in terms of small step semantics. It looks more like a sequent proof system, quite intentionally. 

Possibly `MM2` (with registers `a` and `b`) and which is also standalone, could be moved there as well. 

The question is more why `MM` is not under `CounterMachines` and the reason is because of the quite involved local reasoning framework that is used to show the correctness of the compilers to the `MM` target. Btw, that framework was also used with `BSM` as the target, ie. the reduction from `BPCP` to `BSM`.

Both `MM2` and `ndMM2` can be used as seeds w/o knowing anything about the `MM` local reasoning framework.

mrhaandi:
> The reason is that `ndMM2` is standalone and does not depend on the `MM` framework at all.

This confuses me even more. `ndMM2` also does not depend on the counter machine framework that I developed for the semi-unification results. However, this should not be the point. It should not matter for which particular result the problem was specified, but to which problem class it belongs. If we separate MinskyMachines and CounterMachines (which in itself could be debated, and I'm fine both ways), then I would still argue that both `MM2` and `ndMM2` belong in the same MinskyMachine directory because their specification says that they are such.

DmxLarchey:
> This confuses me even more. `ndMM2` also does not depend on the counter machine framework that I developed for the semi-unification results. However, this should not be the point...

Honestly I thought this would please you specifically... I missed obviously ;-) 

This is fine. The next push is going to move `ndMM2` into the `MinskyMachines` subdir.

DmxLarchey:
For some unknown reason, the `CI` process fails very early in trying to retrieve `https://coq.inria.fr/opam/released`. Is the server down ?

mrhaandi (approved):

DmxLarchey:
@yforster : does this build script need to depend on whether `coq.inria.fr` is up and running ?

DmxLarchey:
Hi @yforster, I would like to merge this PR. There is minimal interaction with the `coq-8.12` master branch except of course in `_CoqProject` ... Best. D.

yforster:
Looks good to me, I'll merge it now

PR #75, opened 2020-11-04
Hoare Framework for TMs in 8.12

fakusb:
(This superseeds #51 )
I had a look into @mwuttke97's hoare framework for TM verification.

Proofs in it are a lot more maintainable than "old" TM, as they don't need auto-generated names. It also leads to much more understandable proof states during TM-verification, especially in case of something going wrong.

I ported it to the current code base and made use of the existing Hoare-verification for the Universal machine.
(In total, this saves compile time (~20 sec) even when accounting for the files of the hoare-framework itself.


fakusb:
I made this a draft, as the printing of hoare-triples could still be improved to hide the garbage not currently relevant.

mwuttke97:
Looks good.

I think last year I also *started* porting LM to Hoare, but then I moved on to other things before comming far. I could not find this code. Porting LM to Hoare would have sped up compilation tremendously.

fakusb:
I close this as it will be included in #72.

PR #76, opened 2020-11-05
Reduction of SBTM halting to SRH, removing singleTM

yforster:
Plus some cleanup work renaming files according to the guidelines.

Should be fairly uncontroversial and can be merged if CI agrees.


mrhaandi:
Very nice cleanup.
The only thing that got me wondering is `Require Import Undecidability.TM.Reductions.HaltSBTM_to_HaltSBTMu.` in `StringRewriting/Reductions/HaltSBTMu_to_SRH.v`. At first glance `HaltSBTM_to_HaltSBTMu` and `HaltSBTMu_to_SRH` should contain atomic reductions and compile in parallel and not necessarily depend on one another.

yforster:
Good catch, the import is not needed.

mrhaandi (approved):

DmxLarchey:
Hi Yannick,

This is a welcome cleanup indeed. I will just make the following comment. I was a bit puzzled at first by the statement of the problem. Yes I know it might have been there for a while and I never looked into it seriously ... 

But indeed, you describe big-step semantics as:

```
Inductive eval (M : SBTM) : state M -> tape -> state M -> tape -> Prop :=
| eval_halt q t :
    trans M (q, (curr t)) = None ->
    eval M q t q t
| eval_step q t q' w m  q'' t' :
    trans M (q, curr t) = Some (q', w, m) ->
    eval M q' (mv m (wr w t)) q'' t' ->
    eval M q t q'' t'.
```

However the undecidability statement does not need the output value. So you could just as well remove the output `q'` and `t'` from the `eval q t q' t'`, and may be call it `terminates_from q t`.

Maybe you need the output value to compose machines or for intermediate results but at the highest level, the output value does not seem to be needed. Or else you also want to consider the computed values at some point for stronger equivalence results ?

Best, Dominique

DmxLarchey (approved):
I approve the PR

PR #77, opened 2020-11-06
Switch to GitHub Actions for CI

yforster:
Travis has changed its pricing model and is now charging after using 16 hours of CI - which is not enough for us by far.

I thus experimented a bit with GitHub actions, and it seems to be working well, including the caching. Big plus: The GitHub CI machines seem to be faster, so we now only need around 21 minutes of CI compilation time.

Once #76 is merged, I'll merge this one as well.


.github/workflows/build.yml
```diff
@@ -0,0 +1,31 @@
...
+          ocaml-version: 4.07.1
+
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 1
+
+      - run: opam repo add coq-released https://coq.inria.fr/opam/released
+      - run: opam repo add psl-opam-repository https://github.com/uds-psl/psl-opam-repository.git 
+      - run: opam install . --deps-only --with-doc --with-test
```
fakusb, line 29: Why "--with-doc --with-tests"? Seems slower

PR #79, opened 2020-11-14
Removed external dependency on PslBase

fakusb:
Removed external dependency on PslBase by including it in theories/Shared/Libs/PSL.
Notably, PslBase is only used in `L`, `TM`, and `HaltSBTM_to_HaltBSM`, so this should not be of any concern for people except maybe @yforster .

This simplifies installation and makes the addition of the psl-opam-repository not necessary (unless one installs the Undecidability library itself via opam).

I also fixed that the Vector notation: "[| ... |]" does not interfere with destruct patterns "destruct x as [|]" anymore

The following things _would_ also nice-to-have, and easier to do now that PslBase is included:
-Some files from Shared and Shared/Libs/PSL are very similar due to historic reasons, and one might can tidy this up.
-On the other hand, the files in "toplevel" of shared are probably meant to never be used again, and only there for other code to work?
-Remove Shared/Prelim.v dependencies to make Andrej more happy
-Libs/PSL declares several hints in core, which changes the behaviour of the tactic "now" and annoys me sometimes

This includes https://github.com/uds-psl/base-library/pull/8, which speeds up Undec-Lib  builds a tiny bit although they now compile a few more files.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,9 +12,9 @@ jobs:
         uses: actions/cache@v2
         with:
           path: "~/.opam"
-          key: opam-${{github.base_ref}}-${{github.ref}} 
+          key: opam-noPslBase-${{github.base_ref}}-${{github.ref}} 
           restore-keys: |
-            opam--refs/heads/${{github.base_ref}}
+            opam-noPslBase--refs/heads/${{github.base_ref}}
 
       - name: Install OCaml
         uses: avsm/setup-ocaml@v1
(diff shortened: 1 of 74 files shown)


fakusb:
In theory, it might be cleaner to move `Shared/Lib` to `Lib` and move the top-level-files of `shared` intro `Lib/Shared`

yforster:
I think it's good to have a library folder. Currently it's named Shared, which is maybe confusing. But I think it's good to have only problem directories on top level, not also the multiple library directories we have.

At some point, we might want to rename Shared into something more reasonable, but it's okay to leave it like this for now I think.

@fakusb can you push a new commit to let CI re-run against the merged PR #74?

@mrhaandi @DmxLarchey feel free to intervene if you think this shouldn't be merged. Otherwise I'd like to merge soon. This also enables us to release an opam package in the official repo, because now all our dependencies are in the official opam.

DmxLarchey:
I am fine with removing PslBase dependency. Also some notations defined in that library interfered badly with my own libs. Overall, I do not think any library should be mandatory everywhere so certainly a file called `Shared/Prelim.v` was not ideal.

Issue #81, opened 2020-11-16
library 1.0.0+8.12 opam release


mrhaandi:
Here is a wish list for the upcoming opam release of the undecidability library version `1.0.0+8.12` (we'll try to hit 20.11.)
@yforster @fakusb @DmxLarchey 
- [x] remove `PslBase` dependency
- [x] refactor/remove `Shared.Prelim`
- [x] compiles with coq 8.12.1
- [x] cleanup output ~~(what is all the `phi1 = ∀ Pred (!P) (Vector.cons (var 0) (Vector.cons (var 0) Vector.nil)) : form` and `prv_class is not universe polymorphic` output?)~~
- [x] fix warnings ~~(I get `Warning: Notation "_ [ _ ]" was already defined with a different format` and `Warning: Not a truly recursive fixpoint.`)~~ ~~By the way, why are those not shown in the CI logs?~~
- [x] move the final FOL files away from `Problems` and `Reductions` ~~(no need to fully clean up FOL now)~~
- [x] paper-ready System F presentation
- [x] Hoare framework rework (I'm not sure about his one, @fakusb knows more)
- [x] cleaner, modular, commented synthetic definitions. Our agreed focus are non-informative reductions. ~~I do not expect (exposed!) notation definitions such as `"⪯ᵢ"` in `ReducibilityFacts.v` that is included everywhere.~~
- [x] remove uncompiled files
- [x] more uniform FOL

By the way, the suggested version `X.Y.Z+A` is:
- `X` increases on major design changes
- `Y` increases on new undecidability results
- `Z` increases on minor improvements
- `A` is the Coq version used

yforster:
I'll fix the CI logs, can take care of the FOL stuff, repair the subst notation and will clean up the output regarding FOL. The fixpoint warning probably has to stay, but we can add it to the ignored warnings in CoqProject.

mrhaandi:
> The fixpoint warning probably has to stay

In both cases the fixpoint warnings could be fixed by 
```
Definition de_bruijn_ext {X : Type} := 
  fix de_bruijn_ext (ν : nat -> X) (x : X) (n : nat) {struct n} : X :=
    match n with
    | 0 => x
    | S n => ν n
    end.
```

In general, I'm against disabling any warnings (even unfixable ones). If warnings completely pollute the output, then clearly the problem lies elsewhere.

yforster:
Is this in the H10 part? My comment assumed that the problem occurs in the L directory.

[...]

PR #82, opened 2020-11-17
Cleanup regarding FOL

yforster:
TODO list:

- [x] move `FOL`-related files from `Problems` and `Reductions
- [x] fix directory structure in `FOL`, add `FOL_undec` file
- [x] move `FOLP`-related files from `Problems` and `Reductions`
- [x] CI run


yforster:
This fixes

- [x ] cleanup output (what is all the `phi1 = ∀ Pred (!P) (Vector.cons (var 0) (Vector.cons (var 0) Vector.nil)) : form` and `prv_class is not universe polymorphic` output?)
- [x ] fix warnings (I get `Warning: Notation "_ [ _ ]" was already defined with a different format` and `Warning: Not a truly recursive fixpoint.`) ~~By the way, why are those not shown in the CI logs?~~
- [x] move the final FOL files away from `Problems` and `Reductions` (no need to clean up FOL now)

from #81, apart from the warnings: There is one notation warning left regarding MM that I could track down, I'll leave that for @DmxLarchey.

Since this is for sure strictly better than the previous state, I'd suggest that I merge this once CI agrees on its last run. Please intervene quickly if you disagree :)

PR #86, opened 2020-11-17
change Fixpoint to Definition ... := fix to avoid a warning

DmxLarchey:
Removes two kinds of _harmless_ warnings:

- `Coq` complaining about `Fixpoint` definitions that make no recursive sub-calls. `Fixpoint`s are somewhat natively better behaved wrt the `simpl` tactic. Solution is to replace by anonymous fixpoints, ie `Fixpoint foo ... := ...` is replaced with `Definition foo := fix foo ... := ...`;
- a component access notation `_ [ _ / _ ]` which what declared twice, once with the default (or absent) `format` directive, and a second time with a specific `format` directive to avoid unnecessary blank spaces, hence the warning.

The component access notation is declared as `Reserved Notation` in `Shared/Libs/DLW/Vec/vec.v` where it covers the case of `pos n` indexed vectors but also in `MinskyMachines/MMenv/env.v` where it covers `X -> Y` where `X` is a discrete type.

### DIFF
--- a/theories/H10/Dio/dio_logic.v
+++ b/theories/H10/Dio/dio_logic.v
@@ -23,7 +23,8 @@ Set Implicit Arguments.
 
 (* Fixpoint instead of Definition because of better unfolding properties *)
 
-Fixpoint de_bruijn_ext {X} (ν : nat -> X) x n { struct n } :=
+Definition de_bruijn_ext {X} (ν : nat -> X) x :=
+  fix dbe n :=
   match n with
     | 0   => x
     | S n => ν n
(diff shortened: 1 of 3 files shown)


mrhaandi:
Will you also take care of `Warning: Notation "_ [ _ / _ ]" was already defined with a different format.` in `MuRec` and `MMenv`?

DmxLarchey:
@mrhaandi the last `commit` should solve the issue. It was a conflicting `format` directive for the component access ` _ [ _ /  _ ]` notation that was at odds with the default one.

yforster:
Is this ready to merge?

DmxLarchey:
Yep I do it, these are very local changes that should have no impact...

PR #90, opened 2020-11-19
Delete uncompiled files

yforster:
Delete all files which are not compiled, using Fabian's script:

```
#!/bin/bash
#List all .v files that do not have .vo files next to them (usefull to test what did not actually get compiled)

for f in $(find ./ -name '*.v');
do
    fv="${f%.*}.vo" ;
    if ! [[ -f $fv ]]; then
        ls "$f" 
    fi
done
```

The only controversial one is maybe `TRAKHTENBROT/summary.v`. @DmxLarchey @dominik-kirst is it okay to delete this file? It's still there in your TRAKHTENBROT tag the paper is linking to. Keeping it on the main branch without adding it to `_CoqProject` means that the file can break without us noticing, because CI only cares about files in the `_CoqProject`.


dominik-kirst:
I have no strong opinion but wouldn't keeping an overview file (not just listing the problems and reductions) be a good idea even in other larger directories? For instance, also for the H10 directory it could be interesting to have a list of the main theorems not directly classified as reductions to allow a second entry point next to the paper.

yforster:
I agree that summary files are great, but they should be part of the paper repository / tag, not the main repo. Maybe for the undecidability library upstream branch, such overviews should be part of the wiki, not the `theories` directory.

mrhaandi:
> I have no strong opinion but wouldn't keeping an overview file (not just listing the problems and reductions) be a good idea even in other larger directories?

I would prefer to have a well-documented (comments and readme) + uniform library structure, so the main results are obvious without (sometimes incoherent, paper-related) additional auxiliary summaries.
If a summary is needed for a paper, then this should be kept orthogonal to the library.

DmxLarchey:
The summary file `TRAKHTENBROT/summary.v` contains only `Print Assumptions`, `Check` or `About` directives and no definitions or proofs. However the reason it is not in the `_CoqProject` file is to avoid `Print Assumptions` burning CPU cycles and slowing down the `CI` task... initially, it was in the `_CoqProject` file.

Ideally, it could be part of a `assumptions` or `summary` target in the `Makefile` but I have no idea how to generate specific targets using `_CoqProject`.

I agree that it is not a problem wrt. to the paper because the paper refers to a `git` `tag`. However, generally having summary files that could be compiled manually or run into an IDE is I think something we should keep. I have no a priori on how to implement such an additional feature but maybe the wiki solves this?

fakusb:
I would have no problem if the ci build an additional target (after `all`) which compiles summary files.
I dislike Check/About/... in the default target not because burning CPU circles, but they clutter the output and make it harder for me to find warnings or other problems.

yforster:
It's not a good idea to have summary files as part of CI, exactly because of the burning CPU cycles issue. I wouldn't feel confident merging something where CI is still running, and CI with lots of summary files might take very long once we have lots of summary files.

My suggestion: We put the summary files in the wiki, where they can be copy-pasted to an IDE. There is no real guarantee that the file still compiles, but the wiki article can mention on which the summary file was last checked to compile. This way it's optional to have summary files for the authors of subprojects and they do not clutter the structure in `theories`.

mrhaandi:
I agree with Fabian on 
> dislike Check/About/... in the default target

and with Yannick on good wiki so that
> not clutter the structure in `theories`

Maybe today we could make a particular design decision for the 1.0.0 release?

DmxLarchey:
@yforster Btw `MinskyMachines/MM/mm_comp_old.v` can go away ...

DmxLarchey:
> Ideally, it could be part of a `assumptions` or `summary` target in the `Makefile` but I have no idea how to generate specific targets using `_CoqProject`.

My suggestion was to create targets that are _not compiled by default_, in particular not compiled by the `CI`, but could be compiled by the user with `make check_assumptions` or `make summary` for instance.

yforster:
Short-term suggestion: We remove summary files for now.

Long-term suggestion: We introduce a `summaries` directory next to `theories`, which contains it's own `_CoqProject` and `Makefile` and e.g. a file `TRAKHTENBROT.v` with lots of `Print Assumption Statements.`

Deal?

DmxLarchey:
Well I would agree with

> Long-term suggestion: We introduce a `summaries` directory next to `theories`, which contains it's own `_CoqProject` and `Makefile` and e.g. a file `TRAKHTENBROT.v` with lots of `Print Assumption Statements.`

I would agree with that while I do not know how the coordination of the two `_CoqProject` files would unfold? Would `Undecidability` be considered as a library for the `summary/_CoqProject`to be viable ?

Anyway I think we have a deal. Could you tag the last commit containing summary files? Maybe this is the sole one?

yforster:
Yes, the `summaries/_CoqProject` would contain a line `-Q ../theories` Undecidability` and the `Makefile` can contain a `make -C ../theories all` call.

I'll tag and merge!

yforster:
Or maybe no need for a publicly visible tag. We can just state here that the last commit containing summary files is

https://github.com/uds-psl/coq-library-undecidability/commit/6fdc1a3b7536f47d24836ae8bb1373b639fd38d5

mrhaandi:
> We introduce a summaries directory next to theories, which contains it's own `_CoqProject` and `Makefile` and e.g. a file `TRAKHTENBROT.v` with lots of `Print Assumption` Statements.

I like this suggestion. It separates the concerns of the library from outside interactions with the library.

PR #94, opened 2020-11-20
added opam update and opam list --installed to CI

mrhaandi:
(draft to inspect CI)
- added opam update and opam list --installed to CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,6 @@ jobs:
       - run: opam repo add coq-released https://coq.inria.fr/opam/released
       - run: opam update
       - run: opam upgrade
-      - run: eval $(opam env)
       - run: opam install . --deps-only --with-doc --with-test
       - run: opam list --installed
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,8 @@ jobs:
           fetch-depth: 1
 
       - run: opam repo add coq-released https://coq.inria.fr/opam/released
+      - run: opam update
       - run: opam install . --deps-only --with-doc --with-test
+      - run: opam list --installed
 
       - run: opam exec -- make -j 2 all TIMED=1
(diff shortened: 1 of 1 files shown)


mrhaandi:
Unfortunately, on `opam list --installed` CI outputs `coq 8.12.0`. So, `opam update` is not sufficient.

yforster:
`opam upgrade`?

mrhaandi:
Several times `opam upgrade` (on my local system) led to an unrecoverable state, so I had to hard reset my switch.
We can try though. If it fails, then we can switch to handpicking versions.

fakusb:
I would always explicitly use upam instal coq-8.12.1, never opam upgrade, to avoid the problems andrej mentioned.

fakusb:
You probably need to change the cache-key in the ci file, for example replace the 'nobase' part with coq8.12

fakusb:
Furthermore, with gihub actions, all our forks run ci as well, so there is no need for "ci test" PR drafts.

mrhaandi:
> Furthermore, with gihub actions, all our forks run ci as well, so there is no need for "ci test" PR drafts.

This is only partially true. On my fork the CI works with Coq 8.12.1 and here for the PR it does not.

fakusb:
Oh, yes, right, because of the caching.

20 Nov 2020 13:34:00 Andrej Dudenhefner <notifications@github.com>:

mrhaandi:
`opam upgrade` worked in this instance and could provide a somewhat consistent alternative for the CI.
I'm not sure about the manual cache keys because they require more manual work the more packages are included.
Still, the worst case inconsistency scenario (that actually happened recently) is that local CI was ok with a PR (on 8.12.0), but the forked CI did not compile (on 8.12.1) because of different `Require Import` behavior.

fakusb:
Ok, after external discussion with andrej, I think we *should* use opam-upgrade in the ci-script:
The environment a new user gets up by following our install-instructions is the one that opam-upgrade should produce, and it is better if we notice that some dependencies were not declared correctly in some .opam-file than if a user notices that.

yforster:
I think the `eval $(opam env)` is not needed. It has no effect for calls to `opam` anyways, and the `make` is prefixed by `opam exec` to be run automatically with the right settings.

When I did the setup I had the impression that `eval $(opam env)` doesn't do anything in GitHub workflows.

PR #95, opened 2020-11-20
Ported FOL undecidability proofs to new syntax

dominik-kirst:
I've ported the CPP'19 undecidability proofs (FOL validity, satisfiability, provability) to the new syntax. This will be the unified base for future FOL undecidability proofs and I plan to at least formulate an FSAT problem equivalent to the Trakhtenbrot formulation so that the main results can be stated in the same framework.


yforster (approved):
Very clean, looks really good!

I'm in favour of merging sooner rather than later to not stack too many PRs.

yforster:
I'd be in favour of merging this soon (/now). @DmxLarchey are you okay with this? We can still improve on it later on of course if needed.

theories/FOL/Util/Syntax.v
```diff
@@ -0,0 +1,516 @@
+(** * First-Order Logic *)
+
+
+From Undecidability.Synthetic Require Export Definitions DecidabilityFacts EnumerabilityFacts ListEnumerabilityFacts ReducibilityFacts.
```
DmxLarchey, line 4: Why Exports ?

DmxLarchey:
Well it is not really code I dived into before. I just have one issue with `Require Export`. Except if the file is not supposed to be used by others, in which case it does not matter, I am not in favor of getting unwanted notations to pop up when I load an external library like `FOL/Utils/Syntax.v`. This is what may happen with `Export`. It can also pollute the `Definition` namespace but this is somewhat less painful.

Is there a need for those exports ?

mrhaandi:
This PR is already a good improvement over the previous code (and probably could be merged as is for the 1.0.0 release).
Overall, `FOL/FOL.v`, `FOLP/FOLFS.v` may need some further attention in future. In particular, looking at the mostly uncommented list of Problems in `FOL.v` which need four imports from Util (?), it is not straightforward to understand what is going on.
For comparison, `SystemF/SysF.v` has to deal with similar issues.

dominik-kirst:
@DmxLarchey There was no particular reason for exports, I've now switched to imports.

@mrhaandi I have now verbalised each problem in FOL.v with a comment, do you think this is enough? I could also copy the main definitions if you think this would be helpful.

mrhaandi:
> I have now verbalised each problem in FOL.v

That already helps. 
The minimal specification idea is aiming to minimize the potential number of lines one needs to survey to understand the specification.
Currently the necessary imports `Syntax` `Deduction` `Tarski` `Kripke` make it almost 1000 loc.
> I could also copy the main definitions
Copying probably will make it worse due to redundancy. E.g. the copied/cmmented out "Syntax as defined in Util/Syntax.v" is already somewhat redundant and prone to in future diverge from the actual definition.
Sometimes it is complicated to untangle minimal specification from infrastructure, so maybe the current state is as good as it gets without a complete redesign.

mrhaandi:
I suggest taking the current state into 1.0.0 and make gradual improvements with time in future (maybe improving upon design guidelines).

DmxLarchey (approved):
I am happy with the update that could be further improved later on. It could be merged into the release once the `CI` works out properly.

mrhaandi:
I reproduced the problem. It seems that currently one cannot both `Require Undecidability.FOL.FOL.` and `Require Undecidability.PCP.PCP_undec.` (because of side effect in the specification).
The same is true with `FOL` and the (very dangerous) line `From Undecidability Require Export CodeTM Copy ChangeAlphabet WriteValue.` in `TM/Code/ProgrammingTools.v`. Specifically, the `Code.v` may be responsible for the FOL trouble.

yforster:
What is dangerous about the line in `ProgrammingTools.v`?

I'll try to reproduce the error as well now, then we can discuss what the best fix is (although probably all fixes will be unsatisfying).

mrhaandi:
> What is dangerous about the line in `ProgrammingTools.v`?

`From Undecidability Require Export Copy` could export the contents of any arbitrary `Copy.v` (not necessarily the right one) anywhere in the library (while `Export` itself is less encapsulated than `Import`). Especially since the contents of the `Copy.v` seems to clash randomly with unrelated files.

I'd suggest for the library to never use `From Undecidability Require Import/Export File.v`, but instead always specify the correct path.

yforster:
Closed by #96

PR #96, opened 2020-11-23
Unified FOL syntax without universe inconsistency

yforster:
This is `coq-8.12.` merged with #95, without the universe inconsistency and with fully qualified Exports in `TM/Code/ProgrammingTools.v`.

The universe inconistency came from the definition

`Definition vec := Vector.t`

This is problematic because `vec` is a constant of type `Type{i} -> nat -> Type@{i}` for a fixed `i`. Having the `i` fixed is a problem, because now using `vec` introduces universe constraints mentioning `i`, which in this case were inconsistent.

Three possibilities to fix this:
- work with `Vector.t` explicitly everywhere (in general, I like this)
- use `Notation vec := Vector.t` and live with the cluttered notation space
- use `Local Notation vec := Vector.t` in every file

I went with the last one here, because it was relatively quick to implement.


DmxLarchey:
> * Notation vec := Vector.t

This is what I did in the `Shared/Libs/DLW/Vec/vec.v` library after being invited to unify my own definition of vectors with the one in the standard library. I remember this working out quite flawlessly in my case.

PR #99, opened 2020-11-26
Improvements to L framework

fakusb:
-Cleaned up Require Exports
-Removed `Defined` where `Qed` sufficed
-improved extraction runtime by reimplementing Lrewrite and therefore changing Lsimpl.
-removed clash between L hoas syntax and app-coercion by making `convert` from HOAS to L exlicit (syntax `[L_HOAS someTerm]`)

### DIFF
--- a/theories/L/Complexity/ResourceMeasures.v
+++ b/theories/L/Complexity/ResourceMeasures.v
@@ -47,7 +47,7 @@ Module Leftmost.
     intros s t R u ? <-. revert s t R u.
     induction k;cbn in *;intros ? ? R ?. congruence. destruct R as [s' [R1 R2]].
     exists (app s' u). firstorder.
-  Defined.
+  Qed.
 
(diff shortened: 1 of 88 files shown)


yforster:
Is extraction really faster now? It seems like CI took slightly longer (but I have no idea how consistent CI timing is).

I'm happy to merge this in any case due to the huge amount of cleanup.

fakusb:
I copied the build log from ci to `time-of-build-(before|after).log` and used `make -f Makefile.coq print-pretty-timed-diff`.
Every file took 15-10% longer, even things i did not touch, except for files where extraction is used, where time speed up by 7 to 40%.

So it is due to ci not being a reliable, absolute benchmark tool.

fakusb:
https://gist.github.com/fakusb/9b3f4fdaf3ad7288ea773877d3848e74

PR #101, opened 2020-12-06
Optimized for compiled interfaces (vos)

mrhaandi:
- optimized library for fast compilation of Coq's compiled interfaces [`vos`](https://coq.inria.fr/refman/practical-tools/coq-commands.html#compiled-interfaces-produced-using-vos)
- `make vos` builds (much faster than just `make`) interface files for the library, which makes it usable quickly in e.g. `vscoq`
- `make vok` checks (maximally parallel) the library for correctness using interface files
- added `CI` for compiled interface verification
- `Set Default Proof Using "Type".` is advised for proper `vos` compilation when using `Section`

Motivated by Karl Palmskog and supported by @fakusb .


fakusb:
We also now use the flambda version of ocaml for the CI.

Note that ocaml 4.07.1+flambda (in CI) with [proper flags](https://coq.discourse.group/t/install-notes-on-ocaml-versions-and-coq-configuration/713/6?u=fakusb) (not in CI) leads to 10-20% faster compilation compared to more recent oaml-versions without flambda/without tweaked flags.

DmxLarchey:
What is the behavior of `Proof.` without the `using` option? 

My guess is that w/o the default setting, it can use any variable or `Let` definition available from the environment, to avoid breaking a lot of proof scripts. Then with `Set Default Proof Using Type`, you disable the above behavior so you have to specify each used hypothesis explicitly. Am I correct?

This in turn allows for a proper type synthesis wrt. to `Section` w/o having to fully check all proofs?

What does happens when you force an hypothesis but this hypothesis is in fact not used in the proof? Does the type generated after type checking the proof assume the hypothesis or not?

fakusb:
`Proof using Type` means 'use only the section variables that appear inside the type of the statement'. So I think your first two paragraphs are true.

If you forget a variable that the body od the proof uses (maybe onlt transitively), then coq complaons when it really cheks the proof and tells you, at 'Qed.', how your 'using' annotation should have looked like.

mrhaandi:
@DmxLarchey I've seen some of your code using the `Let ... Proof. ... Qed.` construction which is probably a Coq bug (see https://github.com/coq/coq/issues/13584 ) and also leads to slow interface compilation.
I had to change `Let` to `Local Lemma` at most crucial places for faster compilation.

DmxLarchey:
> I had to change `Let` to `Local Lemma` at most crucial places for faster compilation.

No problem with that. I already did such things in compilers like eg. in `MuRec/ra_mm_env.v` section `ra_compiler_rec`. A lot of `Let` definitions bloat the environment and slow down `omega` or `lia`.

DmxLarchey:
> yes, mentioning 'to much' means the theorem depends on more than necessary after sections get discharged. So it has assumptions in ita type that are not necessary.

Thanks for your answers. Btw, sometimes you want the "theorems" to be over specified because they have to fit a specific shape to be used as input for a general theorem that applies to that shape and not to a stronger one.

.github/workflows/build.yml
```diff
@@ -1,35 +1,96 @@
...
-
-      - run: opam exec -- make -j 2 all TIMED=1
+    - name: Install OCaml
+      if: steps.cache.outputs.cache-hit != 'true'
+      uses: avsm/setup-ocaml@v1
+      with:
+        ocaml-version: ${{env.OCAML_VERSION}}
+
+    - name: Checkout code
+      if: steps.cache.outputs.cache-hit != 'true'
```
yforster, line 28: Do I understand correctly that the Coq installation now happens in the `build-deps` job, and the `build-full` and `build-fast` jobs use the cache stored by the `build-deps` job to compile (and that's why they fail if restoring fails)? I'm a bit surprised that GH actions don't allow sharing build files without caching.

Makefile
```diff
@@ -1,19 +1,10 @@
 all:
 	export TIMED
-	$(MAKE) -C theories all
+	@+$(MAKE) -C theories all
 
-install:
-	$(MAKE) -C theories install
+force Makefile: ;
```
yforster, line 5: What does this line do? I would have maybe expected `force: Makefile`

theories/L/Datatypes/List/List_basics.v
```diff
@@ -23,7 +23,7 @@ Instance term_map (X Y:Type) (Hx : registered X) (Hy:registered Y): computableTi
 Proof.
   extract.
   solverec. all: unfold c__map; solverec. 
-Defined.
+Defined. (*because other extract*)
```
yforster, line 26: What do those comments mean?

fakusb: I wanted to know why I use defined instead QED

Makefile
fakusb: To not pass down the "Makefile" target.

.github/workflows/build.yml
fakusb: Yes, we use cache for that.

PR #107, opened 2020-12-30
Undecidability of ZF Set Theory

dominik-kirst:
I've reworked my undecidability proof of ZF set theory using the unified FOL syntax. The covered problems are as follows:
- Validity and intuitionistic provability in ZF and Z' (i.e. ZF without separation and replacement) with Skolem symbols, see file ZF.v
- Validity and intuitionistic provability in ZF and Z' (i.e. ZF without separation and replacement) without Skolem symbols, see file minZF.v

The undecidability proofs require models of the respective theories, one way to obtain these models using a description operator on wf trees as in our CPP'18 paper is provided. The models for Z' could already be given just assuming extensionality of classes, as also noted in the CPP'18 paper. Further todos are as follows:
- If dropping the extensionality requirement (syntactic equality is interpreted as Coq equality) in the formulation of validity, the models for Z' can be given without assumptions.
- It should be possible to eliminate usage of the infinity axiom so that the constructive HF set model from Smolka/Stark can be used.
- The elimination of Skolem symbols used for the symbol-free axiomatisation should generalise to arbitrary theories.
- The verified translation from ZF with symbols to symbol-free ZF should easily yield the conservativity of the former over the latter.


theories/FOL/ZF.v
```diff
@@ -0,0 +1,155 @@
+(** * Definition of semantic and deductive ZF-Entailment *)
```
yforster, line 1: The `coqdoc` output of the library currently is minimal, but works.

Please either remove `(**` in the development, or make sure that the output is in line with the rest

theories/FOL/ZF.v
```diff
@@ -0,0 +1,155 @@
...
+
```
yforster, line 2: To be accessible, this file would need more comments explaining what's going on. But I wouldn't consider this necessary, just nice to have.

theories/FOL/ZF.v
```diff
@@ -0,0 +1,155 @@
...
+Definition fun_rel phi :=
+  ∀ ∀ ∀ phi[$2 .: $1 .: Nat.add 3 >> var] --> phi[$2 .: $0 .: Nat.add 3 >> var] --> $1 ≡ $0.
+
+Definition ax_rep phi :=
+  fun_rel phi --> ∀ ∃ ∀ $0 ∈ $1 <--> ∃ $0 ∈ $3 ∧ phi[$0 .: $1 .: Nat.add 4 >> var].
+
+Definition ZF' :=
+  ax_ext :: ax_eset :: ax_pair :: ax_union :: ax_power :: ax_om1 :: ax_om2 :: nil.
+
+Inductive ZF : form -> Prop :=
```
yforster, line 103: I don't understand this predicate

yforster (changes requested):
Only the coqdoc output is a showstopper for me, up to that I'm happy to merge

theories/FOL/ZF.v
dominik-kirst: Good point, will do :)

theories/FOL/ZF.v
dominik-kirst: You mean ZF? That's just adding all instances of separation and replacement to the core axioms in the list ZF'.

theories/FOL/ZF.v
dominik-kirst: Hmm is the plan to keep the coqdoc minimal with a single entry each problem and undecidability proof? Or is it okay to have one top-level entry for FOL with a more fine-grained internal structure (roughly one sub-entry each file)? I think the latter would be the preferable structure for publication linking.

theories/FOL/ZF.v
yforster: Okay, but what's `ZF'` then? Is the point to define the first-order theory of all axioms of `ZF`?

(I think the answer to the latter question is "yes". After understanding that `ZF` is supposed to be a theory, I also missed that it has to consider axiom schemes, but understood that now (that's why it is inductive rather than another list). This point might be an important thing to explain in the comments.)

dominik-kirst: Exactly, the strategy is to isolate the minimal (finite) axiomatisation expressible enough to verify the reduction. Then the reduction lifts generically to all (possibly infinite) extensions that still have a standard model, one instance is ZF by adding both separation and replacement, another one could be Z by just adding separation.

mrhaandi (changes requested):
While none of the below comments are absolute blockers, I would encourrage their consideration.

Minor Comments

- Why is this PR against `uds-psl:coq-8.12` and not `uds-psl:coq-8.13`, which will soon be the main branch?
- Some files have inconsistent/redundant trailing newlines; e.g. `PCPb_to_ZFD.v` 672--675 compared to `PCPb_to_ZF.v` no new lines before EOF.
Personally, I prefer exactly one trailing new line before EOF (from other programming languages); but at least be consistent.

Major Comments

- I would discourage (what to me seem) auxiliary lemmas in `_undec.v` files such as `Lemma extensional_eq_min'` in `minZF_undec.v`.
After all, the `_undec.v` files are supposed to contain exactly the (possibly conditional) undecidability results.
- Problem specification is still not quite minimal.
For example, `ZF.v` relies on `FullDeduction` which contains
auxiliary lemmas (not necessary for problem definition), hints to `core` (I would discourage side effects in specification), ltac code (again, not necessary and somewhat intransparent for a surveyor).
I encourage separation of a minimal specification from any auxiliary structures that are used in proofs.
- In a reduction file, e.g. `Reductions/PCPb_to_ZF.v` I would expect something like `Theorem reduction : (condition ->) PCPb ⪯ entailment_ZF.` (cf. `CFG/Reductions/PCP_to_CFPI.v`). This way, it is easier for a third party to use this many-one reduction step (e.g. in a reduction chain, cf. `SemiUnification/Reductions/HaltTM_1_chain_SemiU.v`) in a uniform way.

dominik-kirst:
> * Why is this PR against `uds-psl:coq-8.12` and not `uds-psl:coq-8.13`, which will soon be the main branch?

The PR was filed before the 8.13 release.

> * Some files have inconsistent/redundant trailing newlines; e.g. `PCPb_to_ZFD.v` 672--675 compared to `PCPb_to_ZF.v` no new lines before EOF.
>   Personally, I prefer exactly one trailing new line before EOF (from other programming languages); but at least be consistent.

I wasn't aware of this guide line and I would be actually surprised if this has been consistent so far, at least in any code I have contributed.
 
> * I would discourage (what to me seem) auxiliary lemmas in `_undec.v` files such as `Lemma extensional_eq_min'` in `minZF_undec.v`.
>   After all, the `_undec.v` files are supposed to contain exactly the (possibly conditional) undecidability results.

Agreed, I will move these lemmas.

> * Problem specification is still not quite minimal.
>   For example, `ZF.v` relies on `FullDeduction` which contains
>   auxiliary lemmas (not necessary for problem definition), hints to `core` (I would discourage side effects in specification), ltac code (again, not necessary and somewhat intransparent for a surveyor).
>   I encourage separation of a minimal specification from any auxiliary structures that are used in proofs.

I'm not sure a self-contained spec in a single file is the right way to go here. The problems depend on more than the deduction systems, at least also the syntax and semantics. If I copy the bare definitions into the specification, I'm not sure this will be much more instructive and if I do not copy but move the definitions, I introduce weird dependencies.

I could separate the files like FullDeduction.v into respective spec and facts files though and only import the former in the problem file. Or I make sure that FullDeduction.v is clearly separated in a spec and a facts part. How would you feed about this?

> * In a reduction file, e.g. `Reductions/PCPb_to_ZF.v` I would expect something like `Theorem reduction : (condition ->) PCPb ⪯ entailment_ZF.` (cf. `CFG/Reductions/PCP_to_CFPI.v`). This way, it is easier for a third party to use this many-one reduction step (e.g. in a reduction chain, cf. `SemiUnification/Reductions/HaltTM_1_chain_SemiU.v`) in a uniform way.

Okay good point, I'll state these reduction facts.

mrhaandi:
> Or I make sure that `FullDeduction.v` is clearly separated in a spec and a facts part. How would you feed about this?

Yes, for surveyability this would already be a good improvement.
Would it be possible upon importing the specification not to add any hints to `core` (but maybe to a dedicated FOL hint database if necessary)? Several times in the past using the library I encountered incoherent `eauto` behavior because specifications added `core` hints.

yforster:
As a quick fix, having the hints in `FullDeduction_facts.v` only and none in the spec file would already help, no?

What I don't like about hint databases is that they make calls to eauto syntactically longer. If there would be a way to say "in this file, use the following databases" that would be great.

mrhaandi:
> As a quick fix, having the hints in `FullDeduction_facts.v` only and none in the spec file would already help, no?

Yes, unless they become "weird dependencies" which Dominik mentioned.

> What I don't like about hint databases is that they make calls to eauto syntactically longer. If there would be a way to say "in this file, use the following databases" that would be great.

I sometimes write `eauto with problem_db nocore` for maximal robustness (and sometimes speed).
This already saved some maintenance work (especially moving towards Coq 8.13).
Setting local a default for individual files sound like a nice Coq feature request.

dominik-kirst:
I have just pushed the following changes:
- Coqdoc structure is disabled, I'll enable it in my fork again for the paper submission
- The problem files contain more explanatory comments and only import the definitions (`Syntax.v` 130 lines, `FullTarski.v` 100 lines, `FullDeduction.v` 50 lines). The facts are now in separate `_facts.v` files. If this structure is acceptable, I'll refactor the other FOL undec devs accordingly on the next occasion.
- The `_undec.v` files are now free of intermediate lemmas and I state the reduction facts explicitly (e.g. `PCPb_entailment_ZF` in `ZF_undec.v`). I'd rather keep them in the `_undec.v` files than moving them to the respective reduction files since they also connect the proofs involving the side conditions.

mrhaandi:
> I have just pushed the following changes: [...]

This looks very good. One small thing: `Notation "x ≡ y"` in `ZF.v` (among others) is on top-level (not opt-in) and has no scope. This might cause collisions in future with other `x ≡ y` notations (compare to encapsulation via `Module ListAutomationNotations.` in `Shared/ListAutomation.v`).
Arguably, if `ZF.v` forever remains a leaf, it is not a problem. Still, maybe a `Module ZFNotations` encapsulation would be more future proof.

dominik-kirst:
> This looks very good. One small thing: `Notation "x ≡ y"` in `ZF.v` (among others) is on top-level (not opt-in) and has no scope. This might cause collisions in future with other `x ≡ y` notations (compare to encapsulation via `Module ListAutomationNotations.` in `Shared/ListAutomation.v`).
> Arguably, if `ZF.v` forever remains a leaf, it is not a problem. Still, maybe a `Module ZFNotations` encapsulation would be more future proof.

The syntax notations in `ZF.v` and `minZF.v` are now in a new scope called `syn`. The same symbols are reused as semantic notations in a scope called `sem` that I already introduced before.

mrhaandi (approved):

yforster (approved):

theories/FOL/ZF_undec.v
```diff
@@ -0,0 +1,93 @@
...
+
+Corollary undecidable_model_entailment_ZF' :
+  inhabited extensional_normaliser -> undecidable entailment_ZF'.
+Proof.
+  intros (V & M & H1 & H2 & H3) % normaliser_model. apply undecidable_entailment_ZF'.
+  exists V, M. split; trivial. split; trivial. eauto using ZF.
+Qed.
+
+Theorem PCPb_deduction_ZF :
+  (exists V (M : interp V), extensional M /\ standard M /\ forall rho psi, ZF psi -> rho ⊨ psi) -> PCPb ⪯ deduction_ZF.
```
DmxLarchey, line 46: Dear @mrhaandi. Do we allow conditional reduction statements now?

theories/FOL/ZF_undec.v
```diff
@@ -0,0 +1,93 @@
...
+Qed.
+
+Theorem undecidable_entailment_ZF' :
+  (exists V (M : interp V), extensional M /\ standard M /\ forall rho psi, In psi ZF' -> rho ⊨ psi) -> undecidable entailment_ZF'.
+Proof.
+   intros H. now apply (undecidability_from_reducibility PCPb_undec), PCPb_entailment_ZF'.
+Qed.
+
+Corollary undecidable_model_entailment_ZF :
+  inhabited extensional_normaliser -> undecidable entailment_ZF.
```
DmxLarchey, line 33: Dear @dominik-kirst. It is quite not obvious to unfold the definition of what is an `extensional_normaliser`? Could you please give an informal description in comment at least?

DmxLarchey:
Well it seems I am too late ...

theories/FOL/ZF_undec.v
mrhaandi: A more precise question is either 
1) is the above condition an inherent part of the problem statement and should actually be presented via a subtype (e.g. assume Turing machines have at most 3 states), or
2) is this a meta-theoretical condition (e.g. assume excluded middle holds)
In case 1) the theorem should not be conditional (the condition has to be part of problem specification); in case 2) conditional statements are fine.
I'm not an expert for `PCPb_deduction_ZF` and since Dominik's PR is clearly improving the previous code I did not strongly object.

yforster:
It's never too late to improve :) We can discuss the questions and then do a new PR improving the current state. But I was willing to merge early to enable Dominik's "this work was contributed to the Coq Library of Undecidability Proofs" sentence in the paper draft and since the PR contributed an interesting and new result.

DmxLarchey:
> We can discuss the questions and then do a new PR improving the current state.

This is fine with me ;-)

dominik-kirst:
@DmxLarchey I'll make sure these points are addressed in the next PR.

Indeed, the condition is necessary to strengthen Coq enough to contain a model of ZF, since a reduction from a non-trivial problem to an axiomatisation proves the axiomatisation consistent. So it is option (2) of Andrej's list.

Informally, the assumption yields an extensional unique choice operator on isomorphism classes of well-founded trees, so it can be seen as a strong quotient axiom. There are 2 reasons why the computability of the reduction function should not be affected by this assumption. First, the reduction is defined in an axiom-free context before a ZF model is assumed to verify it. Secondly, not the consistency of CT but its admissibility is relevant for the actual computability of defined functions and our understanding is that any consistent _propositional_ assumption maintains admissibility of CT. In fact, I don't think that such a tree normaliser is inconsistent with CT since it only introduces a modest amount of unique choice and no classical logical principle like EM or LPO.

That said, I plan to weaken the assumption in the case of the considered fragments (ZF without replacement, ZF without any axiom scheme) because for them extensionality of classes (following from FE and PE) is enough to get a model. Also, the first two todos mentioned in the initial PR message will completely dispose of assumptions, since lifting the extensionality requirement (still without replacement) and eliminating the infinity axiom both admit fully constructive models.

PR #111, opened 2021-01-26
Peano Arithmetic, Binary ZF, Binary FOL

dominik-kirst:
This PR contains the following problems shown undecidable:
- Semantic Tarski entailment and intuitionistic natural deduction in Peano arithmetic (`FOL/PA.v`), contributed by Marc Hermes (@Lereau)
- An alternate formulation of Diophantine equations without parameters (`H10/H10p.v`) as a simpler starting point for PA
- ZF over the binary signature containing only membership (`FOL/binZF.v`), similar to the already merged formulation with membership and equality (`FOL/minZF.v`)
- Validity and intuitionistic provability in FOL over the binary signature (`FOL/binFOL.v`) using the finite fragment of binary ZF

Moreover, the assumptions for the undecidability of ZF problems have been refined as mentioned in previous discussions:
- Extensional validity in the fragment axiomatisations only requires extensionality of classes (`CE`), before it used unique choice on equivalence classes (`TD`)
- Deduction in the fragment axiomatisations is completely free of assumptions
- An intensional version of validity is given, for which the undecidability proofs are also free of assumptions


mrhaandi (changes requested):
My main issue is again the content of `_undec.v` files for which their agreed upon, uniform purpose is to contain exactly the `undecidable PROBLEM` theorems.
Now looking (among others) at `H10p_undec.v` we have some
- `Definition embed_poly ...`
- `Lemma embed_env_eval ...`
- `Theorem H10_H10p : H10 ⪯ H10p_SAT.`

all of which do not belong there. In fact, the last point is a reduction, which should go into a `Reductions` folder. Also, the agreed upon uniform naming for such a theorem would not be `H10_H10p`, but rather `reduction` in the file `H10_to_H10p_SAT.v`.
For comparison, see `PCP/PCP_undec.v` and `PCP/Reductions.v`.
One of the reasons to put atomic `X ⪯ Y` theorems in individual files is (besides uniform organization) not to require `X ⪯ Z` theorems, if all such things were put into one `_undec.v` file as currently the case in e.g. `binFOL_undec.v`.

DmxLarchey:
The way I read it is `H10p` is just a reformulation of `H10` over `nat` for variables instead of `Fin n` (also called `pos n`), `n` being the bound on the number of variables, an extra input value for the description of the problem. I.e. `H10` is _already a parameter-free formulation_ of Hilbert tenth problem.

dominik-kirst:
@DmxLarchey To me it looks like that `H10_Problem` still uses the full syntax of `dio_polynomial` instantiated for the empty parameter type, leaving a dead constructor. This is of course fine but I wanted to provide Marc with the more direct variant as a starting point for his master's thesis and I think the new formulation is the slightly more natural seed problem. If you think this is too redundant, I'm happy to eliminate it from the PR.

dominik-kirst:
@mrhaandi I'll change `H10p_undec.v` with the next commit, I just thought the reduction is short enough to be included in the same file.

The placement of reduction theorems was probably a misunderstanding, in the discussion on my last PR I understood (apparently wrongly) that they should go to the `_undec.v` files. Marc already moved them in the case of PA and I'm fine with that. In the case of ZF I'm not so sure since the reductions depend on assumptions and I feel that the refinement of these assumptions should be made in a central file, otherwise all reduction files depend on the model constructions.

DmxLarchey:
> you think this is too redundant

No no that was not my point. My point was that in your presentation of this PR, you sounded like there was _no parameter free_ version of `H10`. I am perfectly fine with multiple "equivalent" formulations of a given problem to fit the need of flexibility in the search for a convenient seed, whatever "equivalent" means here.

So may be you just replace 

- Diophantine equations without parameters (`H10/H10p.v`) as a simpler starting point for PA 

with 

- An alternate formulation of Diophantine equations without parameters (`H10/H10p.v`) as a simpler starting point for PA.

dominik-kirst:
@DmxLarchey Oh I see, I've changed the PR description as you suggested. I'll also add a respective comment in the `H10/H10p.v` file.

dominik-kirst:
@mrhaandi and @DmxLarchey I've just changed the H10p reduction according to your feedback.

yforster:
> In the case of ZF I'm not so sure since the reductions depend on assumptions and I feel that the refinement of these assumptions should be made in a central file, otherwise all reduction files depend on the model constructions.

I don't understand this point. If the reduction theorems depend on models, for sure the reduction files will depend on files constructing the models. Why would this be a problem?

mrhaandi (approved):

yforster (approved):
That looks very good, with the standard "but":

Please update the README (add the new problems and reductions, and also add Marc to the contributors)

DmxLarchey:
Some random suggestions for improvements:

- May I suggest introducing notations for vectors like eg `x##y##ø` ? This would make the other definitions of notations in `ZF.v` much easier to read. 
- Also use `_` for explicit arguments that can be deduced from later ones eg `ZF_func_sig`.
- Is this comment `(* Formulas have falsity as fixed constant -- we could parametrise against this in principle *)` in `Syntax.v` still relevant ? It seems to me that paramerization has been implemented ?
- In `FullTarski.v` which seems to be about semantics, convenient syntax for FOL is also defined? I would rather see that in `Syntax.v` ? Also, the convenient syntax is not used in the definition of `Fixpoint sat {ff : falsity_flag} (rho : env) (phi : form) : Prop := ...`

DmxLarchey:
Concerning `CE` which is postulated for the semantic undecidability results, it is still a property of the type `Acz`, actually the predicates over `Acz` which is a very large type. Hence, it does not only postulate that there exists some model of ZF but that a particular type is actually enough to build a model of ZF. Am I right ? I still find this assumption very strong.

dominik-kirst:
Thanks @DmxLarchey, that's really helpful! :)

> * May I suggest introducing notations for vectors like eg `x##y##ø` ? This would make the other definitions of notations in `ZF.v` much easier to read.

I could import `VectorNotations` as in `PA.v` if you think they are sufficient.

> * Also use `_` for explicit arguments that can be deduced from later ones eg `ZF_func_sig`.

I did this at some point but then for using the notations in concrete formulas broke.

> * Is this comment `(* Formulas have falsity as fixed constant -- we could parametrise against this in principle *)` in `Syntax.v` still relevant ? It seems to me that paramerization has been implemented ?

You're right, I'll delete it.

> * In `FullTarski.v` which seems to be about semantics, convenient syntax for FOL is also defined? I would rather see that in `Syntax.v` ? Also, the convenient syntax is not used in the definition of `Fixpoint sat {ff : falsity_flag} (rho : env) (phi : form) : Prop := ...`

It's also the operator instance before the notations, I'll se if I can move all instances to the `Syntax.v` file. I can use the notations in `sat` if you prefer.

> * In particular, I do not see a reason for semantics to be required for the definition of _deduction_ predicates based on `prv`.

Agreed.

> * In the notation `Notation "A ⊢ phi" := (prv A phi) (at level 55).` I do not see how Coq is able to guess the `pierce` implicit argument. The `falsity_flag` comes from `A` but for `pierce` I have no clue.

This notations only works in a section with an assumed instance of pierce or if a priority for the two instances is set.

> * What would be required for undecidability of deduction for classical ZF?

The easy way would be LEM, then the deduction system can be shown sound for the standard model. It could also be an A-translation as in the CPP'19 paper suffices to show that if the classical system deduces the solvability, then so does the intuitionistic system. In general, classical deduction is conservative over intuitionistic deduction for all formulas asserting a translated computation to terminate, at least the reduction used in PA has this form.

> * Also _deduction_ is too general to me here. I think you should call it _natural deduction_. One could also consider a Hilbert style proof system instead of _natural deduction_ or any other equivalent proof system?

Will change the PR description. Of course any sound proof system strong enough to derive the reduction formula would work. Since this really needs to be done deeply in the system, ND is just the most convenient formulation.

> * If needed, could a primitive negation operator be defined as a `quantop` ? The identification `~ A = A -> False` is not necessarily obvious in FOL context.

I think reusing quantop for negation wouldn't work since then substitution would treat it as a quantifier. It could of course be introduced as a unary operation, but to my understanding the identification as `A -> False` is fairly standard and therefore was already used in our previous FOL papers.

dominik-kirst:
> Hence, it does not only postulate that there exists some model of ZF but that a particular type is actually enough to build a model of ZF. Am I right ?

Right, therefore the undecidability theorems are mostly formulated in 2 variants, 1 requiring the existence of a model and 1 just assuming CE and using the model construction. Also note that CE does not suffice to prove replacement, for full ZF the description axiom TD seems necessary.

> I still find this assumption very strong. I understand that `CE` is stronger that `PropExt`. Is it equivalent to the following?
> 
> ```
> forall X (P Q : X -> Prop), (forall x, P x <-> Q x) -> P = Q
> ```

It certainly follows from general extensionality of classes that you've formulated (which in turn follows from functionan and propositional extensionality). I don't think that it's equivalent, not every predicate can be encoded as a predicate on well-founded trees, at least in the absence of axioms maximising the `Acz` type.

> Undecidability of _natural deduction_ results are much better behaved. Maybe I repeat myself but assuming external axioms is not harmless to the synthetic method IMHO.
> 
> _Could it be possible to build a non-extensional model of ZF?_ Obviously w/o assuming `CE` but then you would have an explicit class equivalence predicate instead of equality ?

This is exactly done for the intensional problems, for instance `entailment_ZFeq'` in `ZF.v` and the intensional model could even be used for Z additionally assuming separation. The problems for natural deduction behave exactly the same since one direction is via soundness in the standard model, I've just directly used the intensional model since the equality axioms are already assumed for the deductive problems.

So yes, sacrificing extensionality admits models up to Z. If you want to go further to ZF instead of the more predicative CZF, I think there's no way around extending Coq's type theory, both for deductive and semantic problems. On the other hand, if you go further below Z, for instance by eliminating infinity as I think should be possible, you can even use the extensionalised HF model. In total I would vote to keep both intensional and extensional reductions in the library, the former for the more instructive explanation without setoid issues, the latter for the more economical result.

DmxLarchey:
Another question @dominik-kirst. Did you consider the _regularity axiom_ which is a classical formulation of the well-foundedness of membership? 

After a quick try, I can show that `Ain` is constructivelly  `wellfounded` (ie `forall s, Acc Ain s`) but do you think that the regularity axiom holds in the `Acz` model ?

dominik-kirst:
Regularity is usually equivalent to LEM, so it shouldn't be provable for `Ain`. In first-order set theory one would have to adopt a scheme of epsilon-induction for every first-order predicate to ensure well-foundedness without enforcing classical logic. This is discussed in Myhill's initial attempt to define intuitionistic set theory.

DmxLarchey:
Well modulo the small updates discussed above that I guess @dominik-kirst is going to integrate into this PR, I do approve it.

DmxLarchey (approved):

PR #120, opened 2021-04-06
Improvements in the proof of Trakhtenbrot

DmxLarchey:
Not supposed to be merged right now. More precise description, to be completed.

+ Improved notations to correspond more closely to the paper (LMCS)
+ a new finite quotient along a partial equivalence relation and a list of defined elements (not necessarily all of them) in `Shared/Libs/DLW/Utils/fin_quotient.v`
+ `list_seteq` in `Shared/Libs/DLW/Utils/seteq.v`:
  - a nice inductive characterization of extensional identity of lists (same members)
  - `list_seteq` is permutation augmented with contraction equivalence
  - used to simplify the proof of `bti_equiv` in `btree.v`.
+ better automation in `btree.v`, most proofs are now very short.
+ cleanup (removed dead code) and better notations in `membership.v`
+ simplified the reduction from `Σn` to `Σ2` by removing some unnecessary axioms, impacts `reln_hfs.v` and `Sign_Sig2.v`
+ a better tactic notation `fol equiv *` for proving `A o B <-> A' o B'` where `o` is a binary connective or other. Not all the code is fully updated to use this tactic. 

A finer characterization for decidability of `FSAT` over monadic signatures (arity less than one):
+ as in the `discrete` signature case, one has to distinguish if there is a unary relation or else if all relation symbols are nullary;
+ two symbols are _discernable_ if they can be mapped to two different values in `bool`, eg `f x = true` and `f y = false` for some `f : rels -> bool` (equivalently, they can be mapped to two different values of a discrete type);
+ if there is a unary relation symbol: _`FSAT` is decidable iff discernability of function symbols and of relation symbols are decidable_;
+ if all relation symbols are nullary (arity 0):  _`FSAT` is decidable iff discernability of relation symbols is decidable_.

### DIFF
--- a/theories/Shared/Libs/DLW/Utils/fin_dec.v
+++ b/theories/Shared/Libs/DLW/Utils/fin_dec.v
@@ -10,9 +10,19 @@
 Require Import List Arith Lia.
 
 From Undecidability.Shared.Libs.DLW.Utils
-  Require Import fin_base.
+  Require Import utils_decidable fin_base.
 
 Set Implicit Arguments.
+
+Theorem finite_t_find_dec X (P : X -> Prop) 
+           (Pdec : forall x, { P x } + { ~ P x }) 
+           (HQ : finite_t X) :
+           { x | P x } + { forall x, ~ P x }.
+Proof.
+  destruct HQ as (l & Hl). 
+  destruct list_choose_dep with (P := P) (Q := fun x => ~ P x) (l := l)
+    as [ (? & ? & ?) | ]; eauto.
+Qed.
 
 Theorem exists_dec_fin_t X (P Q : X -> Prop) 
            (Pdec : forall x, { P x } + { ~ P x }) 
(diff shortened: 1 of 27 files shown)


dominik-kirst:
Shouldn't this PR go against the `coq-8.13` branch as the soon-to-be default branch?

DmxLarchey:
> Shouldn't this PR go against the `coq-8.13` branch as the soon-to-be default branch?

Ok I will try to merge it with `coq-8.13` ... but is it becoming `default` anytime soon ?

dominik-kirst:
> Ok I will try to merge it with `coq-8.13` ... but is it becoming `default` anytime soon ?

That's my understanding, that's why I based the separation logic mechanisation on `coq-8.13`. If I remember correctly, @yforster just needs someone to ping him? ;-)

DmxLarchey:
Ok @dominik-kirst. I merged with `coq-8.13` and created PR #123

yforster:
Got the ping, now `coq-8.13` is the default branch. I initially wanted to only switch branches once there is an `opam` package for `8.13`. But given the activity at the moment I'd guess it makes more sense to first merge all the open PRs and then have a look at packaging again.

DmxLarchey:
Hi @yforster Yannick,

Since you got the ping, I also ping you on PR #126 for `coq-8.13` as well ;-)

PR #122, opened 2021-04-21
Improvements (notations, shorter proofs, etc) for IMSELL

DmxLarchey:
This PR provides improvements to the `IMSELL` undecidability results code. I would like it to be merged quickly so that I can refer to it in the final version of the upcoming FSCD21 paper [pre-print](https://members.loria.fr/DLarchey/files/papers/fscd21.pdf) that presents this work and also the _previously undocumented_ reduction from `FRACTRAN` to `MMA(2)` and `MMA0(2)` (two counters alternate Minsky machines). 

Also some updates to the library code:

+ improved `subcode_tac` in `Shared/Libs/DLW/Code/subcode.v` to solve more subcode detection automatically. Some proofs get a bit simplified in the `MM` and `BSM` code, ie. `subcode_tac; ...` is removed on many occasions in eg `StackMachines/BSM/bsm_utils.v` and `MuRec/ra_mm_env.v`;
+ better notations for alternate `MM` (ie. `INCₐ` instead of `INC`, but they are synonymous) to remind of the alternate semantics;
+ some cleanup of proofs in `FRACTRAN`.

Although some of my shared libs have been slightly improved, and some cleanup in the `FRACTRAN` code, this PR should not impact the rest of the project.

I also intend to create a _tag on a fork of this commit_ with a specialized `README.md` that would look like [this one](https://github.com/DmxLarchey/coq-library-undecidability/tree/fscd21). Fortunately, tags do not pollute the branch namespace ;-)

### DIFF
--- a/theories/MinskyMachines/MM.v
+++ b/theories/MinskyMachines/MM.v
@@ -46,6 +46,10 @@ Inductive mm_instr (X : Set) : Set :=
 Notation INC := mm_inc.
 Notation DEC := mm_dec.
 
+Notation INCₐ := mm_inc.
+Notation DECₐ := mm_dec.
+
+
 (* ** Semantics for MM, restricted to X = pos n for some n *)
 
 Section Minsky_Machine.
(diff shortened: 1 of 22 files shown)


theories/MinskyMachines/MM.v
```diff
@@ -46,6 +46,10 @@ Inductive mm_instr (X : Set) : Set :=
 Notation INC := mm_inc.
 Notation DEC := mm_dec.
 
+Notation INCₐ := mm_inc.
```
mrhaandi, line 49: I understand your reasoning, if one already knows the other model.
From a modular perspective, i.e. treating each problem as an individual unit, this is confusing.
The change is ok, although I wouldn't have made it.

mrhaandi (approved):

mrhaandi:
By the way, why is this PR for the `8.12` branch and not the `8.13` one?

DmxLarchey:
> By the way, why is this PR for the `8.12` branch and not the `8.13` one?

Ok I will propose a PR for `8.13` very quickly.

DmxLarchey:
I may dump that PR if PR #126 goes into `coq-8.13`.

PR #123, opened 2021-04-22
Improvements in the proof of Trakhtenbrot (for coq-8.13 branch)

DmxLarchey:
Not supposed to be merged right now. More precise description, to be completed.

+ Improved notations to correspond more closely to the paper (LMCS)
+ a new finite quotient along a partial equivalence relation and a list of defined elements (not necessarily all of them) in `Shared/Libs/DLW/Utils/fin_quotient.v`
+ `list_seteq` in `Shared/Libs/DLW/Utils/seteq.v`:
  - a nice inductive characterization of extensional identity of lists (same members)
  - `list_seteq` is permutation augmented with contraction equivalence
  - used to simplify the proof of `bti_equiv` in `btree.v`.
+ better automation in `btree.v`, most proofs are now very short.
+ cleanup (removed dead code) and better notations in `membership.v`
+ simplified the reduction from `Σn` to `Σ2` by removing some unnecessary axioms, impacts `reln_hfs.v` and `Sign_Sig2.v`
+ a better tactic notation `fol equiv *` for proving `A o B <-> A' o B'` where `o` is a binary connective or other. Not all the code is fully updated to use this tactic. 

A finer characterization for decidability of `FSAT` over monadic signatures (arity less than one):
+ as in the `discrete` signature case, one has to distinguish if there is a unary relation or else if all relation symbols are nullary;
+ two symbols are _discernable_ if they can be mapped to two different values in `bool`, eg `f x = true` and `f y = false` for some `f : rels -> bool` (equivalently, they can be mapped to two different values of a discrete type);
+ if there is a unary relation symbol: _`FSAT` is decidable iff discernability of function symbols and of relation symbols are decidable_;
+ if all relation symbols are nullary (arity 0):  _`FSAT` is decidable iff discernability of relation symbols is decidable_.


DmxLarchey:
Hi @dominik-kirst. I did improve on your comments on the paper. Did you like the development of _discernability_ ?

dominik-kirst:
> Hi @dominik-kirst. I did improve on your comments on the paper. Did you like the development of _discernability_ or does it feel like splitting hairs (the french wording for that is rude)?

I just pushed some more replies, looks pretty good to me now :)

The section on discernability is a great fit to the paper, since it once again showcases a subtlety of constructive finite model theory that is completely invisible for the classical view. The fine structure of `discreteness -> dec apartness -> dec discernability` is very interesting, especially since it contributes to a more precise signature classification in comparison to the IJCAR paper. If the reviewers don't like it we can still shorten, but for now I would just keep it as is.

By the way I'm now convinced that discernability is really a weaker property than apartness, even on strongly finite types. There can always be hidden intensional properties showing x apart from y that are not computably detectable. So in particular think that the `dec apartness -> dec discernability` direction is strict.

DmxLarchey:
I did push some remarks on the paper. Could you reference the code in the paper (sources files and name of results or definitions) ?

DmxLarchey:
> By the way I'm now convinced that discernability is really a weaker property than apartness, even on strongly finite types.

After thinking a bit about it, I realized that decidable discernability is a necessary condition for SAT over decidable models (`P` and `Q` are discernable iff there is a model of `P /\ ~ Q`) whereas discreteness is a necessary condition for PROVABILITY (`P` and `Q` are equal iff `P -> Q` is provable).

dominik-kirst:
> I did push some remarks on the paper. Could you reference the code in the paper (sources files and name of results or definitions) ?

In the separation logic section? Can do, I guess we'll merge the developments at some point so I set the links accordingly?

DmxLarchey:
> In the separation logic section? Can do, I guess we'll merge the developments at some point so I set the links accordingly?

Well with the PR gets merged, we are ok. Isn't the sep-logic devel. already in `coq-8.13` ?

DmxLarchey:
> > `P` and `Q` are equal iff `P -> Q` is provable
> 
> How do you prove the backwards direction without arguing semantically?

In a system with cut-elimination, I would say this entails there is a cut-free proof of `P |- Q` and then
by inversion of the rules (excluding cut).

dominik-kirst:
> Well with the PR gets merged, we are ok. Isn't the sep-logic devel. already in `coq-8.13` ?

Yep it is. So I've now added preliminary links and as soon as we have merged and reuploaded the coqdoc all links should work.

dominik-kirst:
> In a system with cut-elimination, I would say this entails there is a cut-free proof of `P |- Q` and then by inversion of the rules (excluding cut). May be I forgot to precise that `P` and `Q` are variables.
> 
> For the semantic argument, soundness is enough (no need for completeness) but the model would need to computably differentiate `P` from `Q`.

Okay, I see this :)

I think the underlying fact is that discernability corresponds to decidable models and discreteness to arbitrary models (which in turn corresponds to provability). In the latter case also the semantic argument works since there is no need to differentiate computably.

DmxLarchey:
About the _Boolean counterpart_? Does it correspond to [`squash`](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.13/theories/SeparationLogic/Reductions/FSATdc_to_MSLSAT.v) ? Is there any advantage in projecting a decidable proposition to `if dec P then True else False` instead of `if dec P then true else false = true` ? Otherwise it seems to be the same construction.

dominik-kirst:
> About the _Boolean counterpart_? Does it correspond to [`squash`](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.13/theories/SeparationLogic/Reductions/FSATdc_to_MSLSAT.v) ? Is there any advantage in projecting a decidable proposition to `if dec P then True else False` instead of `if dec P then true else false = true` ? Otherwise it seems to be the same construction.

Ah, I thought that you use an explicit Boolean membership function. Of both versions the `if dec P then True else False` is slightly easier to show unique, at least without using lemmas.

DmxLarchey:
> Ah, I thought that you use an explicit Boolean membership function. Of both versions the `if dec P then True else False` is slightly easier to show unique, at least without using lemmas.

Oh no, I just _squashed_ at the last possible time but on `bool` instead of `True/False`. I agree you do not need to show `UIP` for `@eq bool` with your version.

DmxLarchey:
> I've just regenerated the HTML files from you development branch and pushed it to the new website so we can check if the new links work.

The paper links seem to work fine. However:
* the [TOC](https://www.ps.uni-saarland.de/extras/fol-trakh-ext/website/toc.html) is not up to date and does not reflect the contents of the paper
* the [summary file](https://www.ps.uni-saarland.de/extras/fol-trakh-ext/website/Undecidability.TRAKHTENBROT.summary.html) is missing

These are the two main pointers so it is a bit annoying.

> Btw, when do we plan to submit? Technically, the deadline is tomorrow AoE, do you think this will now be strict? I think I can work on the remaining todos in the intro later today, what else would be missing for submission?

I do not think much is missing now for the submission. I would have liked the paper to link to a GH `tag` of the main library. We can submit and generate the tag later thought. I would like to customize the `README.md` in the given tag to explain what to compile (eg `make TRAK.../red_undec.vo`) or reintroduce the `summary.v` file.

dominik-kirst:
> The paper links seem to work fine. However:
> 
> * the [TOC](https://www.ps.uni-saarland.de/extras/fol-trakh-ext/website/toc.html) is not up to date and does not reflect the contents of the paper
> * the [summary file](https://www.ps.uni-saarland.de/extras/fol-trakh-ext/website/Undecidability.TRAKHTENBROT.summary.html) is missing

I've added the old summary file html, seems the links are still intact. I've had the issue with the TOC before, we can't have both the sparse structure needed for the library and the detailed structure following the paper at the same time. What I did is to merge the code at some point and then reintroduce the comments generating the paper structure in my fork. For now I've just deleted the link to the TOC, I'll add it again if we decide to add a meaningful one.

> I do not think much is missing now for the submission. I would have liked the paper to link to a GH `tag` of the main library. We can submit and generate the tag later thought. I would like to customize the `README.md` in the given tag to explain what to compile (eg `make TRAK.../red_undec.vo`) or reintroduce the `summary.v` file.

I think for submission it would be enough to link to your development branch with additional explanations (especially if an older snapshot of it has been merged to the main library).

Should we maybe zoom at some point tomorrow to clarify these questions?

DmxLarchey:
> Should we maybe zoom at some point tomorrow to clarify these questions?

Let us zoom tomorrow. Morning is ok for you ?

I did push some updates to the paper, removing some remarks but also adding some (just to visualise the addtions). Feel free to correct directly on them. What do you think of the para. on constructive subtleties ?

See you tomorrow,

Dominique

dominik-kirst:
> Let us zoom tomorrow. Morning is ok for you ?

Yep, would 10am be fine? The link I've sent you a couple of days ago should still work.

DmxLarchey:
> Yep, would 10am be fine? The link I've sent you a couple of days ago should still work.

Ok for 10am. See you.

dominik-kirst:
I've just pushed my final todos in the tex file. I've also reuploaded the old coqdoc toc but it doesn't really work well (only some of the files jump into meaningful parts). You can give it a try (accessible from the top right of any of the html pages) and then we'll see if we keep it.

DmxLarchey:
Hi @dominik-kirst ,

I loaded and _updated_ the `summary.v` file. The file was not in sync so it took a while...

Could you append the results for `sep_logic` at the end of this file, following the scheme, i.e. `Print` of needed definitions. `About` and `Print Assumptions` for results.

DmxLarchey:
Hi @dominik-kirst, shall I create a new paper on Arxiv or update the existing one? I guess I create a new one but I just wanted to be sure.

dominik-kirst:
> Hi @dominik-kirst, shall I create a new paper on Arxiv or update the existing one? I guess I create a new one but I just wanted to be sure.

Yeah a new one seems right, otherwise the references to the extended IJCAR paper from before will break.

dominik-kirst:
> Could you append the results for `sep_logic` at the end of this file, following the scheme, i.e. `Print` of needed definitions. `About` and `Print Assumptions` for results.

I've added them in my local clone of your fork, but I don't have permission to push it. You either have to give me write access to your fork or I'll just post the few lines here?

DmxLarchey:
@dominik-kirst I assigned you to this PR. Does this give you write access?

dominik-kirst:
Doesn't seem so, I think the PR management is really independent of the repo access. But nevermind, I've anyways just added the following lines, so maybe you can just do the commit:

From Undecidability.SeparationLogic Require Import SL MSL SL_undec Reductions.FSATdc_to_MSLSAT.

Print sp_form. (* Syntax of full separation logic *)
Print msp_form. (* Syntax of fragment *)
Print sp_sat. (* Satisfiability relation *)
Print SLSAT. (* Corresponding decision problem *)

About reduction. (* Reduction from FSAT' to separation logic *)
About SLSAT_undec. (* Undecidability of SLSAT *)

dominik-kirst:
Pushed. Are you fine with this?

dominik-kirst:
Hm the HTML doesn't include the normal (* ... *) comments anymore, I'll see if I can fix that later.

dominik-kirst:
> Hm the HTML doesn't include the normal (* ... *) comments anymore, I'll see if I can fix that later.

Okay fixed.

DmxLarchey:
**Urgent** @dominik-kirst : Seems that Lemma 55 in the paper does point to `FSATdc_to_MSLSAT.reduction_backwards` but the statements are unrelated as I read them...

dominik-kirst:
> **Urgent** @dominik-kirst : Seems that Lemma 55 in the paper does point to `FSATdc_to_MSLSAT.reduction_backwards` but the statements are unrelated as I read them...

Looks fine to me, `FSATdc_to_MSLSAT.reduction_backwards` states the equivalence in both satisfiability relations with the condition on the free variables. The dummy element `(a0, (none, none))` in Lemma 55 is assumed as `d0 : dom h` which looks indeed a bit different. The conditions in the constructed model (finite, discrete, decidable) are then listed in the following lemmas.

DmxLarchey:
This is the Coq output for `About FSATdc_to_MSLSAT.reduction_backwards.` 

```
reduction_backwards :
∀ (s : stack) (h : heap) (d0 : dom h) (phi : Syntax.form),
 (∀ x : nat, x ∊ FV phi -> ∃ l : nat, s x = Some l /\ (l, (None, None)) ∊ h) ->
 FullTarski.sat (env s h d0) phi <-> msp_sat s h (encode phi)

reduction_backwards is not universe polymorphic
Arguments reduction_backwards _ _ _ _ _%function_scope
reduction_backwards is opaque
Expands to: Constant Undecidability.SeparationLogic.Reductions.FSATdc_to_MSLSAT.reduction_backwards
```

I am sorry but I cannot make it match with Lemma 55... In the paper formulation, the existance of the model is above the `<->`, not on its left side.

DmxLarchey:
Also Lemma 55 says _one can compute_ while `FullTarski.sat` is probably `Prop` bounded.

DmxLarchey:
How I read `reduction_backwards` is:
_given a stack s, a heap h and FO formula phi st every free variable in phi is interpreted by a void pointer in h, we have phi is FSAT <-> (s,h) satisfy phi_bar_

DmxLarchey:
Oh may be I see what is my problem: ` FullTarski.sat` is _satisfaction_ not satisfiability !! Is this correct ?

dominik-kirst:
> Oh may be I see what is my problem: ` FullTarski.sat` is _satisfaction_ not satisfiability !! Is this correct ?

Exactly, and the "one can compute" part is the definition of `dom`, `env`, and `model` before that.

> Also the way you are using implicit parameters is confusing me. In fact the _one can compute_ is hidden in the definition of `env` and `env` hides the model as its implicit parameter...

I agree that this is a bit unfortunate. Do you want me to pass around the arguments more explicitly? (As long as you're fine with the way the lemmas are stated in the paper I see no need to worry, we can change the code and coqdoc until we're happy.)

DmxLarchey:
> Doesn't seem so, I think the PR management is really independent of the repo access.

It seems you are not a _maintainer_ of the main repository so I cannot give you the right to push into one of my PR. However I gave you the right to push in my own branch on my fork which automatically updates the corresponding PR, like it was me who did the push to my branch. 

However it does not seem this authorization can be made branch specific. So please do not mess with the other branches of my fork ;-)

I think I do have a maintainer status because I remember I could update a PR once. However, I cannot manage the list of maintainers. I think @yforster is the one who can. On principle, I do not see a reason to limit the updates on PRs to maintainers but this is the annoying state of things.

DmxLarchey:
The branch seems good now. To avoid polluting the main repository with irrelevant comments, I will now create a new PR and later on, remove the two previous PRs (including this one):

- The new PR will contain this code, except for `TRAKHTENBROT/summary.v` and will revert to the official `Readme.md`
- Once accepted by the reviewers, I will branch to a _tag_ containing the `summary.v` and the customized `Readme.md`. This will create a release. I suggest `TRAK-LMCS` a name for the tag/release.
- Then @dominik-kirst can modify the link in http://www.ps.uni-saarland.de/extras/fol-trakh-ext/ to link to the tagged branch.
- Possibly also, he might need to run `coqdoc` once again.

DmxLarchey:
It is possible that the reviewers ask for restructuration of the code along the guidelines. I will write a comment that we need this structure because of the links with the submitted paper and `coqdoc`, and that we intend to create a tag from this PR. But also, we will propose a sub-sequent PR with restructuring a the `TRAKHTENBROT` directory along the guidelines that will hence not interact with the submission.

PR #129, opened 2021-05-22
A summary file for H10 LMCS

DmxLarchey:
To be completed ... this is not intended to be merge in the main branch

### DIFF
--- a/theories/H10/standalone.v
+++ b/theories/H10/standalone.v
@@ -1,0 +1,187 @@
+(**************************************************************)
+(*   Copyright Dominique Larchey-Wendling [*]                 *)
+(*             Yannick Forster            [+]                 *)
+(*                                                            *)
+(*                             [*] Affiliation LORIA -- CNRS  *)
+(*                             [+] Affiliation Saarland Univ. *)
+(**************************************************************)
+(*      This file is distributed under the terms of the       *)
+(*         CeCILL v2 FREE SOFTWARE LICENSE AGREEMENT          *)
+(**************************************************************)
+
+(* Standalone file for "Hilbert's tenth problem in Coq" 
+   by D. Larchey-Wendling and Y. Forster 
+
+   Herein we present a standalone version of the many-one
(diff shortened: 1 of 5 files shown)


DmxLarchey:
The build `build-fast` fails because of some `Print Assumptions` that seem forbidden. That is no problem, this commit is not going to be merged in `coq-8.13`.

DmxLarchey:
Hi @yforster. If you agree with this branch, I will convert into a branch `H10-LMCS-v1.1` and then you can run `coqdoc`. Later on, I will convert it into a release/tag but I wait a bit because changing a release is possible but really not convenient to do.

yforster:
I disabled the `build-fast` target so the red CI won't confuse potential reviewers. Otherwise I'm very happy with this branch!

theories/H10/standalone.v
```diff
@@ -0,0 +1,208 @@
...
+
+(* A polynomial equation e : p = q is a pair (p,q) of poly *)
+Definition H10nat (e : poly*poly) := let (p,q) := e in exists φ, ⟦p⟧ φ = ⟦q⟧ φ.
+
+Section H10_H10nat.
+
+  (* The reduction from 
+       - H10 as defined in the library, unknowns ranging over pos n
+         and parameters over the empty type pos 0
+       - H10nat as defined above, unknows ranging over U/nat
```
DmxLarchey, line 133: unknows type here

theories/H10/standalone.v
```diff
@@ -0,0 +1,208 @@
...
+        | dp_comp do_mul p q => dp2p p ⊗ dp2p q
+      end.
+
+    Variable (φ : pos n -> nat) (ψ : U -> nat).
+
+    Local Definition valuations_match := forall x, φ x = ψ (pos2nat x).
+
+    Hypothesis Hφψ : valuations_match. 
+
+    Fact dp2p_spec p ν : dp_eval φ ν p = ⟦dp2p p⟧ ψ.
```
DmxLarchey, line 155: Should be Local here

PR #133, opened 2021-12-06
Undecidability of SOL and PA2

mark-koch:
The contributed files consist of a mechanisation of the undecidability of validity and satisfiability in second-order logic (SOL.v) and second-order Peano arithmetic (PA2.v). The file Reductions/H10p_to_SOL.v contains the verified reduction from H10p.


README.md
```diff
@@ -199,5 +202,6 @@ We have [guidelines for the directory structure of projects](https://github.com/
 - Simon Spies
 - Dominik Wehr
 - Maximilian Wuttke
+- Mark Koch
```
yforster, line 205: we use alphabetic order here, so you can move yourself up :)

theories/SOL/PA2.v
```diff
@@ -0,0 +1,64 @@
+(** * Second-Order Peano Arithmetic *)
```
yforster, line 1: please check that the coqdoc output is still acceptable. It might be that you have to change all `(**` comments to `(*` to keep the coqdoc clean

theories/SOL/Util/VectorUtil.v
```diff
@@ -0,0 +1,240 @@
...
+  induction v; cbn. reflexivity. intros. f_equal; firstorder.
+Qed.
+
+Lemma In_map {X Y n} (f : X -> Y) (v : vec X n) x :
+  In x v -> In (f x) (map f v).
+Proof.
+  intros H. induction v; cbn. easy. destruct H as [->|]. now left. right. now apply IHv.
+Qed.
+
+Lemma In_map_iff {X Y n} (f : X -> Y) (v : vec X n) y :
```
yforster, line 158: is this and similar basic stuff about vectors not already somewhere in the `Shared` directory? It might be worth moving it there to make it easier for other people to find. I would not expect that I have to check the `SOL` directory to find facts about vectors when searching for them

theories/SOL/SOL.v
```diff
@@ -0,0 +1,195 @@
...
+#[global] Instance empty_preds_sig : preds_signature := {| preds := False; ar_preds := fun P => match P with end |}.
+
+
+(* ** List of decision problems *)
+
+(* Validity of formulas *)
+Definition SOL_valid (phi : form) := forall M rho, @sat _ _ (M_domain M) (M_interp M) rho phi.
+
+(* Satisfiability of formulas *)
+Definition SOL_satis (phi : form) := exists M rho, @sat _ _ (M_domain M) (M_interp M) rho phi.
```
yforster, line 195: this is unusually long for a file defining a problem. Is it possible to split into just problem definition + auxiliary stuff, or is all this really necessary to define the problem?

dominik-kirst: it's indeed all necessary and we were in fact a bit unsure how to proceed. in the case of FOL we had even more code contributing to the problem definitions, so we went with splitting it up over multiple files. now for SOL we have fewer problems (only full Tarski semantics) and so decided it could be at the border to being compact enough for a single file. I think @mrhaandi had a strong opinion that problem files be self-contained as long as they're not as extensive as in FOL?

README.md
mark-koch: Done ^^

theories/SOL/PA2.v
mark-koch: Ok, I have turned every `(**` into `(*` to be save

theories/SOL/Util/VectorUtil.v
mark-koch: Good point, I removed the duplicate lemmas. The remaining ones only concern the `Forall` predicate on vectors. I moved those to a new file `Shared/Libs/PSL/Vectors/VectorForall.v`. Is that ok?

theories/SOL/SOL.v
```diff
@@ -0,0 +1,195 @@
...
+    | quant_pred : quantop -> nat -> form -> form.
+
+End Syntax.
+
+Arguments var_func {_} _ {_}, {_} _ _.
+Arguments func {_ _}.
+Arguments var_pred {_} _ {_}, {_} _ _.
+Arguments atom {_ _ _ _}.
+
+Notation "$ x" := (var_indi x) (at level 5, format "$ '/' x").
```
mrhaandi, line 68: Since these notation cannot be local, I suggest to encapsulate notations in a Module (similar to `ListNotations`). Otherwise, they may not accidentally clash on some transitive inclusion.

theories/SOL/SOL.v
mrhaandi: For me it is not trivial to understand the current specification. For example, how does the lemma `Instance scons_env_func domain ar` contribute to either `Definition SOL_valid` or `Definition SOL_satis`? (similar for some other instances)

yforster (approved):
I'm happy now :) Once the others agree I'll merge.

dominik-kirst (approved):

theories/SOL/SOL.v
mark-koch: @mrhaandi I refactored and added more comments to the code regarding type classes. I hope the specification is a bit clearer now?

mrhaandi: Yes.

mrhaandi (approved):

theories/Shared/Libs/PSL/Vectors/VectorForall.v
```diff
@@ -0,0 +1,121 @@
+(* * Forall operations on ttors *)
+
+Require Import Vector Lia.
+Require Import Undecidability.Shared.Dec.
+From Undecidability.Shared.Libs.PSL Require Import Vectors.
+
+
+Definition Forall {X} (p : X -> Prop) := fix Forall {n} (v : t X n) :=
```
DmxLarchey, line 8: In the [`Shared/Libs/DLW/Vec/pos.v`](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.14/theories/Shared/Libs/DLW/Vec/pos.v) and [`Shared/Libs/DLW/Vec/vec.v`](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.14/theories/Shared/Libs/DLW/Vec/vec.v) libraries, all these _finitary_ quantifications are implemented using `forall p : pos n, ...` where `pos := Fin.t` is the finite type of _positions_. Also used in the `TRAKHTENBROT` devel.

Although these quantifications are convenient for lists as they avoid manipulating proofs that indices do not overflow, they are overkill for vectors, IMHO. This is a side remark, not a request to change anything.

DmxLarchey (approved):

PR #134, opened 2021-12-10
Port everything to 8.15

yforster:
Everything compiles apart from `L` because MetaCoq does not yet have an `8.15` branch. `smpl` has an `8.15` branch. `coq-equations` not yet, but the relevant commit from which one can install manually is in the CI config.

Main changes:
- `apply lem with (x0 := ...)` is now fixed to be `apply lem with (x := ...)`
- the behaviour of `/` in `Arguments` changed, which affected `Vectors.v` in `PSL` and subsequently all vector automation tactics in `TM`
- `Require Import Wf.` is now ambiguous and has to be `Require Import Init.Wf.`
- using `at "2"` in a tactic notation breaks Coq's parser (https://github.com/coq/coq/issues/15318). I temporarily changed the tactic to be `"at" "two"`. The following commands should revert the fixes once Coq is fixed: 
   ```
   git sed "s/POP one/POP 1/g"
  git sed "s/POP zero/POP 0/g"
  git sed "s/at two/at 2/g"
  git sed "s/at three/at 3/g"
  git sed "s/DEC zero/DEC 0/g"
  git sed "s/\"zero\"/\"0\"/g"
  git sed "s/\"one\"/\"1\"/g"
  git sed "s/\"two\"/\"2\"/g"
  git sed "s/\"three\"/\"3\"/g"
  ```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,8 +38,8 @@ jobs:
         opam update
         opam upgrade
 #        opam pin add -n -y -k git coq-metacoq-template.dev+8.14 "https://github.com/MetaCoq/metacoq.git#3d83286"
-        opam pin add -n -y -k git coq-smpl.8.15 "https://github.com/uds-psl/smpl.git#54a001b79
-        opam pin add -n -y -k git coq-equations.dev+8.15 "github.com/mattam82/Coq-Equations.git#722cccfe0
+        opam pin add -n -y -k git coq-smpl.8.15 "https://github.com/uds-psl/smpl.git#54a001b79"
+        opam pin add -n -y -k git coq-equations.dev+8.15 "github.com/mattam82/Coq-Equations.git#722cccfe0"
         opam install . --deps-only --with-doc --with-test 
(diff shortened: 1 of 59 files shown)


yforster:
One more change:
- before, `specialize (H x1 ... xn) as H2` behaved like `pose proof (H x1 ... xn) as H2`. Now, it behaves like `specialize (H x1 ... xn)`; rename H into H2`, so I changed this command to `pose proof`

I also refactored the PSL library a little, we now have
```
$ make TM/TM.vo
coq_makefile -f _CoqProject -o Makefile.coq
make[1]: Entering directory '/home/yannick/research/coq-library-undecidability/theories'
COQDEP VFILES
COQC Shared/Libs/PSL/EqDecDef.v
COQC Shared/Libs/PSL/FiniteTypes/FinTypesDef.v
COQC TM/TM.v
```

yforster:
Re. Warnings there is only one new one that affected us: `Typeclasses Opaque / Transparent` now also needs an explicit locality

PR #136, opened 2021-12-15
Undecidability by Co-Enumerability of HaltL

### LINKED ISSUE #119: Definition of undecidability
Dominik has  brought up that the definition of undecidability we use is too strong, since some problems cannot be proved undecidable constructively. In particular, we can currently not prove synthetically that the _complement_ of the halting problem is undecidable, but since some proofs of undecidability (like satisfiability of FOL) are by reduction from the complement of the halting problem. This is unsatisfying, because one can constructively show that the complement of the halting problem is non-synthetically undecidable if one uses a model of computation.

The easy way out is to define `undec p := decidable p -> enumerable (compl Halt)`. This is also in line with the CPP '19 on synthetic undecidability (with the Entscheidungsproblem as example), where the undecidability axiom `UA` is introduced as `~ enumerable (compl PCP)` (and we of course have `enumerable (compl Halt) <-> enumerable (compl PCP)`. We would have `undec_old p ->undec p`, where `undec_old` is our current definition, so all current proofs are still valid.

As mentioned by Andrej, we could also use semi-decidability instead of enumerability in the definition. Since they are equivalent I'd advocate for choosing enumerability because I find synthetic enumerability easier to explain than synthetic semi-decidability.

I'm opening this issue so we can discuss this centrally and everybody can share their thoughts.

mrhaandi:
- define `undecidable p` as `decidable p -> enumerable (complement HaltL)` in `Undecidability.v`
- establish `undecidable (complement HaltL)` in `L_undec.v` (not possible with previous definition)
- establish `undecidable (HaltL)` in `L_undec.v`
- establish `enumerable (HaltL)` in `L_enum.v` (not possible with `HaltTM 1`)

The initial seed problem `HaltL` is chosen because (without axioms) `HaltTM 1` (also its domain) is not enumerable.
This PR prepares for the joint work with @dominik-kirst and @JoJoDeveloping regarding small FOL fragments.

Closes #119

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -38,8 +38,8 @@ jobs:
         opam update
         opam upgrade
 #        opam pin add -n -y -k git coq-metacoq-template.dev+8.14 "https://github.com/MetaCoq/metacoq.git#3d83286"
-        opam pin add -n -y -k git coq-smpl.8.15 "https://github.com/uds-psl/smpl.git#54a001b79
-        opam pin add -n -y -k git coq-equations.dev+8.15 "github.com/mattam82/Coq-Equations.git#722cccfe0
+        opam pin add -n -y -k git coq-smpl.8.15 "https://github.com/uds-psl/smpl.git#54a001b79"
+        opam pin add -n -y -k git coq-equations.dev+8.15 "github.com/mattam82/Coq-Equations.git#722cccfe0"
         opam install . --deps-only --with-doc --with-test 
         opam list --installed
   
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -29,7 +29,7 @@ jobs:
       uses: actions/checkout@v2
       with:
         fetch-depth: 1
-    
+
     - name: Build dependencies
       if: steps.cache.outputs.cache-hit != 'true'
       run: |
@@ -37,9 +37,9 @@ jobs:
         opam repo add coq-core-dev https://coq.inria.fr/opam/core-dev
         opam update
         opam upgrade
-#        opam pin add -n -y -k git coq-metacoq-template.dev+8.14 "https://github.com/MetaCoq/metacoq.git#3d83286"
+        opam pin add -n -y -k git coq-metacoq-template.dev+8.15 "https://github.com/yforster/template-coq.git#dad61c22c3"
         opam pin add -n -y -k git coq-smpl.8.15 "https://github.com/uds-psl/smpl.git#54a001b79"
-        opam pin add -n -y -k git coq-equations.dev+8.15 "github.com/mattam82/Coq-Equations.git#722cccfe0"
+        opam pin add -n -y -k git coq-equations.dev+8.15 "https://github.com/yforster/Coq-Equations.git#b70c4acf8"
         opam install . --deps-only --with-doc --with-test 
         opam list --installed
   
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 env:
   OCAML_VERSION: 4.07.1+flambda
-  CACHE_KEY: opam-flambda-8.14-newmetacoq-${{github.base_ref}}-${{github.ref}} 
+  CACHE_KEY: opam-flambda-8.15-newmetacoq-${{github.base_ref}}-${{github.ref}} 
 jobs:
   build-deps:
     runs-on: ubuntu-latest
@@ -16,7 +16,7 @@ jobs:
         path: "~/.opam"
         key: ${{env.CACHE_KEY}}
         restore-keys: |
-          opam-flambda-8.14-gitnewmetacoq--refs/heads/${{github.base_ref}}
+          opam-flambda-8.15-gitnewmetacoq--refs/heads/${{github.base_ref}}
 
     - name: Install OCaml
       if: steps.cache.outputs.cache-hit != 'true'
--- a/README.md
+++ b/README.md
@@ -7,8 +7,9 @@
 The Coq Library of Undecidability Proofs contains mechanised reductions to establish undecidability results in Coq.
 The undecidability proofs are based on a synthetic approach to undecidability. 
 A problem `P` is considered [undecidable](theories/Synthetic/Undecidability.v#L8) if its [decidability](theories/Synthetic/Definitions.v#L15) in Coq implies the [enumerability](theories/Synthetic/Definitions.v#L24) of the complement of halting problem for the call-by-value lambda-calculus (`HaltL` in [`L/L.v`](theories/L/L.v)).
+Since the halting problem for the call-by-value lambda-calculus is enumerable (`HaltL_enum` in [`L/L_enum.v`](theories/L/L_enum.v)), enumerability of its complement would classically imply its decidability.
 
-As in the traditional literature, undecidability of a problem `P` in the library is often established by constructing a [many-one reduction](theories/Synthetic/Definitions.v#L27) from an undecidable problem to `P`.
+As in the traditional literature, undecidability of a problem `P` in the library is often established by constructing a [many-one reduction](theories/Synthetic/Definitions.v#L40) from an undecidable problem to `P`.
 
 For more information on the structure of the library, the synthetic approach, and included problems see [Publications](#publications) below, our [Wiki](https://github.com/uds-psl/coq-library-undecidability/wiki), look at the [slides](https://www.ps.uni-saarland.de/~forster/downloads/slides_coqpl20.pdf) or the [recording](https://www.youtube.com/watch?v=mo_C6664n3E) of the talk on the Coq Library of Undecidability proofs at [CoqPL '20](https://popl20.sigplan.org/details/CoqPL-2020-papers/5/A-Coq-Library-of-Undecidable-Problems).
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 The Coq Library of Undecidability Proofs contains mechanised reductions to establish undecidability results in Coq.
 The undecidability proofs are based on a synthetic approach to undecidability. 
-A problem `P` is considered [undecidable](theories/Synthetic/Undecidability.v#L4) if its [decidability](theories/Synthetic/Definitions.v#L15) in Coq implies the [enumerability](theories/Synthetic/Definitions.v#L24) of the complement of halting problem for the call-by-value lambda-calculus (`HaltL` in [`L/L.v`](theories/L/L.v)).
+A problem `P` is considered [undecidable](theories/Synthetic/Undecidability.v#L8) if its [decidability](theories/Synthetic/Definitions.v#L15) in Coq implies the [enumerability](theories/Synthetic/Definitions.v#L24) of the complement of halting problem for the call-by-value lambda-calculus (`HaltL` in [`L/L.v`](theories/L/L.v)).
 
 As in the traditional literature, undecidability of a problem `P` in the library is often established by constructing a [many-one reduction](theories/Synthetic/Definitions.v#L27) from an undecidable problem to `P`.
 
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 
 The Coq Library of Undecidability Proofs contains mechanised reductions to establish undecidability results in Coq.
 The undecidability proofs are based on a synthetic approach to undecidability. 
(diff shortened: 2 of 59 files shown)


DmxLarchey:
Hi @mrhaandi,

So if I get it well, it would not be equivalent to pick up `HaltTM 1`? Can you list which of the main existing seeds would be equivalent to `HaltL` in that respect? We could then point out that any of those seeds would not impact the `undecidabie` (new def.) results.

mrhaandi:
> So if I get it well, it would not be equivalent to pick up `HaltTM 1`? Can you list which of the main existing seeds would be equivalent to `HaltL` in that respect? We could then point out that any of those seeds would not impact the `undecidabie` (new def.) results.

Problematic for `HaltTM 1` are finite functions (needs funext) and the class of finite types (needs some strong axiom for enumerability).
Any seed that is many-one equivalent to `HaltTM 1`, has an enumerable domain, and is enumerable would work well.
For example, `MM2_HALTING`, `PCP`, `FRACTRAN_HALTING`, `RA_UNIV_HALT`, `H10C_SAT` (I'm not sure anymore which many-one cycles we did close).
Of course, one can formulate `HaltTM 1` differently (with tables instead of finite functions, `nat` instead of `Type` for finite types).
We discussed with @yforster and came to the conclusion that it would be too convoluted.

Since `p < q -> ~p < ~q` and we have both `undecidable (complement HaltL)` and `undecidable (HaltL)` we can show undecidability of complements in out library if needed.
The fact `enumerable (HaltL)` connects the approach even stronger with traditional literature.
Since we do not have an axiom free `enumerable (HaltTM 1)` its faithfulness is hand-wavy.

yforster:
I agree with everything Andrej said :) I think it's good to mention that we could equivalently choose other defs, maybe as comment next to the def? Overall, our definition of undecidability now becomes simpler in my opinion: The definition of `L` is a lot lot lot easier to grasp than the long definition of Turing machines.

There is a  downside I want to mention for choosing  `L`: e.g. the undecidability proof of `PCP` (`H10`, etc.) compiles longer, because it depends on the simulation of `L` on Turing machines. But I think it's worth it anyways.

mrhaandi:
> I agree with everything Andrej said :) I think it's good to mention that we could equivalently choose other defs, maybe as comment next to the def? 
I added an explanation together with the two most relevant references to the definition.

DmxLarchey:
> There is a downside I want to mention for choosing `L`: e.g. the undecidability proof of `PCP` (`H10`, etc.) compiles longer, because it depends on the simulation of `L` on Turing machines. But I think it's worth it anyways.

Could a more direct reduction from `L` to `PCP` (or another simpler problem), _not going through Turing machines_ be conceivable, feasible ? Via `Minsky` or `Fractran` or `H10` ?

Would converting `TM` to vectors instead of functions do the trick? I mean not the whole code, just a translation layer showing
`HaltTM 1` <-> `HaltTM_vec 1` and then  `HaltTM_vec 1` would have enumerable co-domain and we could start from that, rewriting just a reduction to `SRS` (I do not remember which one comes first after `HaltTM`).

yforster:
> > There is a downside I want to mention for choosing `L`: e.g. the undecidability proof of `PCP` (`H10`, etc.) compiles longer, because it depends on the simulation of `L` on Turing machines. But I think it's worth it anyways.

> Could a more direct reduction from `L` to `PCP` (or another simpler problem), _not going through Turing machines_ be conceivable, feasible ? Via `Minsky` or `Fractran` or `H10` ?

I think going from `L` to `Minsky` would be possible, but hard. You worked on that a long time ago if I remember correctly :) If we would to that, to go to `PCP` we would still need a simulation of say MM as SRS. I think it makes more sense then to just stay with the proofs we have. (Which, on a higher level, stay relevant anyways because they can also be used as arguments for time complexity)

> Would converting `TM` to vectors instead of functions do the trick? I mean not the whole code, just a translation layer showing `HaltTM 1` <-> `HaltTM_vec 1` and then `HaltTM_vec 1` would have enumerable co-domain and we could start from that, rewriting just a reduction to `SRS` (I do not remember which one comes first after `HaltTM`).

One has to replace `states : finType` by `states : nat` and then use `Fin.t states` everywhere (like in `SBTM`) and replace `trans` by a vector, yes. We used to have something like this a long time ago for an old proof simulating `TM` in `L`. My reasoning why I dislike this (definitely possible) alternative is that we then end up with a formalisation of Turing machines which needs justification. For `L`, we have justification in lots of published papers, same for the current def of Turing machines. For the new one, we'd have no paper to refer to, so it would be a little questionable whether we indeed have a faithful formalisation of Turing machines. And, after all, in my opinion `L` is the nicest model we have, so it seems natural to make it more central.

DmxLarchey:
Hi @yforster,

I think there are two aspects, one "theoretical" and one "practical":

* on the theoretical side, I agree that we want to ground the definition of undecidability on something that can be explained in just a few line, which is the case for `L`, for `PCP` for `H10` or `FRACTRAN` possibly also variants of `MM2`. There may be other ones I forgot. Then of course, having solid literature stating that the chosen problem is justified as a universal model is important too.
* if I get it well, the fact that a reduction chain passes through problems not having enumerable co-domain is _not an issue_ because the existence of the reduction is enough to ensure safe transferring of the definition of undecidability.
* on the practical side, there is the compilation time of the Turing machine code that we want to avoid when we distribute tailored code for a specific paper. For instance, with @dominik-kirst, we reduced `TRAKHTENBROT` from `BPCP` and this compiles in 2-3mins, not 30mins. However, if we want to state `undecidable(TRAKHTENBROT)`, then compiling Turing code is possibly mandatory (but does not require the `L` to `TM` reduction). 
* I have the impression that in the current situation, starting from `L` would involve compiling the `L` to `TM` reduction whenever we want a full proof of `undecidable(something)`.

I am not against making `L` more central, but I would like the center to be closer to the important seeds in terms of compile time. Clearly, the reduction `L` to `TM` is an achievement but then `TM` was the _real_ seed. Marginalizing `TM` could be good but then it would be better to not have it on most importants loops.

DmxLarchey:
Dear @yforster and @mrhaandi,

To progress on this issue of moving `TM` a bit further away from the center, I could suggest the following approach that popped up in my mind this morning. I remember that when we initially started working on the library with @yforster , I was recollecting some formalisation of a compiler from `EXPR` to `BSM` via `FAM`:

* `EXPR` is a functional language with general recursion (similar to C language) but where functions have a unique argument of type binary tree. Programs are given by:
  - a vector of equations `fₙ(x) := eₙ[f₁,...,fₖ](x)` for `n ∈ [1,k]` and `eₙ : expr`;
  - expressions are `expr : e ::= x | ø | ⟨e₁,e₂⟩ | match e with ø => e₁ or ⟨x,y⟩ => e₂ | call fₙ(x)`;
  - an initial expression to evaluate; 
* `FAM` is a functional abstract machine in which `EXPR` programs are compiled into. These are stack machines with two stacks, one for data containing binary trees, and one for return addresses containing natural numbers (or program labels);
* `BSM` are our usual binary stack machines already included in the library.

The compilers `EXPR` to `FAM` and `FAM` to `BSM` already existed in some form, and would just need some cleanup. 

Digging the code, I see that I already wrote a certified `L` evaluator in `EXPR` that was compiled into a 12k instructions `BSM` machine. Notice that writing an `L` CBV evaluator in `EXPR` was not complicated. In the existing code, part of the data is represented by functions (ie environments associating variable names to values) but now that we have a good vector library, I do not seem a difficulty converting these to _extensional_ finite functions that form a datatype, hence enumerable.

So this could be a way to get `HaltTM` out of the loop, or let just say another way of closing the loop. But `BPCP` and `SRS` would also be out of it. Still an issue. In fact I have the impression that the main issue with marginalizing `TM` is keeping at least `BPCP` (or `SRS`) inside the main loop. What do you think?

Would it be possible to simulate `BSM` with `SRS` in an easy way? I do not know precisely how it was done from `TM` to `SRS` ? What are the easy reductions **to** `SRS` or `BPCP`?

DmxLarchey:
> @DmxLarchey If I understand you correctly, you would like to create more, shorter paths in the library. This would be beneficial for small, (mostly) stand-alone reductions from L to PROBLEM (that keep `TM` machinery out). This would solve the practical problem in the particular case and the current PR is good as it is.

Well yes indeed I think more paths let you choose more freely a reduction chain. It also documents other ways to proceed if the library should one day serve as a basis of CS lessons or a book.

In practice, a chain can be longer in steps, but still shorter to compile (because of // compilation). Also longer chains can be easier to explain. Btw I am more and more convinced that the purpose of formalization is not only to check proofs with computers, but to document and explain informal proofs at different levels of granularity.

> Which historically validates `L` as seed.

I have no problem _in principle_ with that choice. Just that the current path from `L` to `BPCP` takes long to compile because of the `L` to `TM` reduction.

> Different approach: From a purely practical perspective, I would avoid in a stand-alone result the `TM` machinery as follows. Copy just the definitions (here it is good that `L` is simple) and as assumtion that the problem which is reduced from is undecidable. If I would like to avoid the axiom, then I would use the library as a library (`opam install`). In any case, such a development would compile really fast and be compatible with the library.

I am not sure I properly get what you suggest here?

dominik-kirst:
> I am not sure I properly get what you suggest here?

I think Andrej's proposal is to sidestep the compilation problem by providing self-contained developments with an assumption for the undecidability of the seed problem. So if we submit an undecidability proof of problem P by reduction from Q, we add the axiom that Q is undecidable and copy the relevant synthetic definitions and lemmas. Then the development is quick to compile and a reviewer who wants to dig deeper can install the library and eliminate the axiom, following instructions we provide.

I like this proposal since it allows for meaningful and compact stand-alone project developments, that are still canonical to merge into the library. And also this somewhat relatives the choice of the main seed, which will always remain pretty arbitrary or driven by practical considerations.

By the way, we're planning to file another PR for Johannes' binary FOL reductions soon, which would depend on the weaker definition of undecidability. Shall we wait until this PR is closed or is it better to already file the new PR using a hypothetical merge of this PR so you can start reviewing?

DmxLarchey:
@mrhaandi Can we merge this PR while keeping the discussion alive? Maybe transfer it to a new PR?

The problem I see with postulating `undec Q` is that when you `Print Assumptions`, now `undec Q` is one. It is just here to avoid including a long chain of reductions but `Print Assumptions` does not tell you that. What I mean is that it would be wrong if `undec Q` was used to implement a reduction.

yforster:
> @mrhaandi Can we merge this PR while keeping the discussion alive? Maybe transfer it to a new PR?

I'm in favour of merging this PR for now so Johannes and others are unblocked. We can continue the discussing on an issue on long compile times then :)

> @dominik-kirst The problem I see with postulating `undec Q` is that when you `Print Assumptions`, now `undec Q` is one. It is just here to avoid including a long chain of reductions but `Print Assumptions` does not tell you that. What I mean is that it would be wrong if `undec Q` was used to implement a reduction.

Off the top of my head: Would it really be problematic to allow theorems of the form `undec PCP -> PCP <= MyProblem`? I'm not arguing that we should allow such reductions (it really doesn't make sense to assume `undec anyhting` to verify a reduction), but they also don't pose a problem to consistency. Under a suitable axiom (`CT`) we can still prove that `~ decidable MyProblem` follows.

yforster (approved):

mrhaandi:
> @mrhaandi Can we merge this PR while keeping the discussion alive? Maybe transfer it to a new PR?
> 
> @dominik-kirst The problem I see with postulating `undec Q` is that when you `Print Assumptions`, now `undec Q` is one. It is just here to avoid including a long chain of reductions but `Print Assumptions` does not tell you that. What I mean is that it would be wrong if `undec Q` was used to implement a reduction.

This discussion is mentioned in #119 which can be kept (and not closed) in order to continue.

@yforster

> theorems of the form `undec PCP -> PCP <= MyProblem`

I do not understand the statement. Should it be `undec PCP -> undec MyProblem` or `HaltL <= PCP -> HaltL <= MyProblem`?

yforster:
> > theorems of the form `undec PCP -> PCP <= MyProblem`
> 
> I do not understand the statement. Should it be `undec PCP -> undec MyProblem` or `HaltL <= PCP -> HaltL <= MyProblem`?

No, the statement is right. I was answering to

> What I mean is that it would be wrong if undec Q was used to implement a reduction.

It would be perfectly fine to assume `undec PCP` when implementing and verifying a reduction - although, probably, it never ever eases anything to do so.

dominik-kirst (approved):

PR #137, opened 2022-01-11
Add undecidability of FOL problems over binary signature and small logical fragment

JoJoDeveloping:
Validity, provability, finite satisfiability and other problems in first-order logic are undecidable as soon as the there is a single binary predicate in the signature.

This has already mostly been mechanized in this library, using different approaches, see [here for a mechanization based on modeling ZF in FOL](https://www.ps.uni-saarland.de/Publications/details/KirstHermes:2021:Synthetic.html) or [here for finite satisfiability by signature compression](https://www.ps.uni-saarland.de/Publications/details/KirstLarchey-Wendling:2020:Trakhtenbrot.html).

We propose new proof of the undecidability of
* validity
* Kripke validity
* satisfiability (†)
* Kripke satisfiability (†)
* provability
* classical provability (assuming XM)
* finite satisfiability (†)
* finite validity (†¹)

by reducing from a variant of Diophantine constraints. This allows us to show that all of these problems are undecidable in the minimal (i.e. single binary predicate) signature and in a minimal logical fragment, consisting of just forall quantifiers, implication, and sometimes negation (for problems marked with †).

This is joint work with @dominik-kirst and @mrhaandi.

1) negation should not be necessary, but we unfortunately could not remove it in our reduction.


theories/DiophantineConstraints/H10C_undec.v
```diff
@@ -48,3 +62,37 @@ Proof.
...
   exact H10SQC_SAT_to_H10UC_SAT.reduction.
 Qed.
+
+Theorem H10UPC_SAT_undec : undecidable H10UPC_SAT.
+Proof.
+  apply (undecidability_from_reducibility H10UC_SAT_undec).
+  exact H10UC_SAT_to_H10UPC_SAT.reduction.
+Qed.
+
+Theorem H10UPC_SAT_compl_undec : undecidable (complement H10UPC_SAT).
```
yforster, line 72: This is strange. Can't you use something like `H10C_SAT_undec` rather than copying the proof for the whole chain?

theories/FOL/Reductions/H10UPC_to_FOL_minimal.v
```diff
@@ -0,0 +1,1036 @@
...
+    end.
+
+  Lemma sig_is_minimal : minimalSignature sig_empty sig_binary.
+  Proof.
+  split.
+  * intros [].
+  * now exists tt.
+  Qed.
+
+  Definition reductionToMinimal (f:falsity_flag) {X:Type} (H:X -> Prop) (P:@form sig_empty sig_binary frag_operators f -> Prop) :=
```
yforster, line 986: [nonblocking] What's the point of having this notation? Wouldn't `H ⪯ @valid sig_empty sig_binary frag_operators falsity_off` be more transparent?

theories/DiophantineConstraints/H10C_undec.v
yforster: That would especially require less `Import`s.

DmxLarchey:
@JoJoDeveloping does it also work with a single binary function symbol and a unary predicate symbol ?

theories/DiophantineConstraints/H10C_undec.v
JoJoDeveloping: Yes, this could be made less hacky by adding a `compl_undec` Lemma for each intermediary problem in this reduction chain.
Shall I do this?

Directly using `H10C_SAT_undec` does not work as `undecidable P -> undecidable (complement P)` does not hold.

JoJoDeveloping:
> does it also work with a single binary function symbol and a unary predicate symbol ?

I conjecture it should work without major reworks, but we have not mechanized it.

theories/DiophantineConstraints/H10C_undec.v
yforster: Oh, I see! In the future we might want to rewrite theorems to talk about the complement. Since your result is the first result relying on the weaker def of undecidability I think it is okay to go with the long proof as is without changing anything

yforster (approved):

theories/FOL/Reductions/H10UPC_to_FOL_minimal.v
JoJoDeveloping: Yes, it would be.

dominik-kirst (approved):

mrhaandi (approved):

PR #139, opened 2022-02-04
Faster vos & no Equations

mrhaandi:
This PR improves maintainability in two ways.
- more rigorous `Set Default Proof Using "Type".` when using sections 
  - faster `vos` compilation
- no explicit use of `Equations` package
  - more controlled `intuition` which uses `auto with *` and behaves differently when `Equations` is present somewhere
  - most of the code compiles without dependencies (only `TM` and `L` use `smpl` and `MetaCoq`)
  - arguably easier to port to new Coq versions
  - most uses of `depelim` were just `Vector.caseS` lemma applications anyway


DmxLarchey (approved):
I did not review every files changed but most of them. I let the biggest to those concerned by the changes. 

I agree with decreasing the degree of dependencies. 

May be we should formalize some guidelines for writing code, not mandatory ones, but recommendations. About avoidable dependencies, about inversions or dependent eliminations, especially wrt stability of the code. See my upcoming comment about those.

DmxLarchey:
First a general remark here. I don't question our very liberal policy about coding style designed to favor of contributions from outsiders. But I do not see any disadvantage about discussing or perhaps publishing a file with code guidelines or tips that everyone could possibly contribute to.  

Concerning `ìnversion` or `depelim`, I usually try to encapsulate their use in tailored lemmas or even better, `Tactic Notations`. See for instance `invert pos`or `vec split` in the libraries [`pos.v`](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.15/theories/Shared/Libs/DLW/Vec/pos.v) and [`vec.v`](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.15/theories/Shared/Libs/DLW/Vec/vec.v).

The reason is that these general inversion tactics while somewhat powerful, generate names in some uncontrollable way (see below). Ideally, I would just use them to write tailored inversion lemmas and nowhere else. So if they break, this happens just at one single point, and in a _context so small that there is an obvious fix_, even for someone that did not write the code in the first place. Also, if you write a general inversion lemma, you might not even need `inversion` at all. It is a combination of `induction` and `match`. Interesting exercise to understand how inversion works.

Some problems that appear when using eg `inversion`. If a name introduced implicitly by `inversion` is used afterwards, the proof script might break:
- if the tactic is updated and the code does not generate the same name; 
- or if the code before is updated and request another name introduced by `inversion` which displaces all the other names;
- also such updates might request new names that become unavailable after their invocation and the code following might break for that reason, eg  `inversion H .... intro H1` where `inversion H` requests `H1` because `H0`has been newly requested by code before.

yforster:
Just a very quick remark: the inversion tactic allows patterns! I didn't know that for a long time, but it's very useful. (unfortunately, this is not yet true for depelim)

DmxLarchey:
> Just a very quick remark: the inversion tactic allows patterns! I didn't know that for a long time, but it's very useful. (unfortunately, this is not yet true for depelim)

I know but the last time I checked, these patterns did not work !! ie the name you put in do not cover all those introduced and also, how do you know what is introduced by inversion if you have a complicated identity for instance.

mrhaandi:
I added more notes to the Wiki https://github.com/uds-psl/coq-library-undecidability/wiki/Structure-Guidelines.
Maybe a more rigorous contribution guidelines are in order.
By the way, in the last years I (almost?) never used `inversion`, but wrote custom inversion lemmas (which is common for `ssreflect`-style proofs).

Maybe one could also localize even further the use of `MetaCoq` and `smpl` to have a more streamlined update process between Coq versions.

yforster:
I read through this discussion again, but I feel I'm missing something so I don't understand the pretense of it.

Do we have a maintainability problem? Sure, maintainability can always be improved a little. But is there an actual problem and we need to devise a plan? The only past problems I am aware of are related to breaking changes in Coq, and we could around the maintainability issues related to that by just adding the library to Coq's CI, which was suggested multiple times to us.

Do we have a problem with dependencies? I'm aware that the slow release cycle of MetaCoq is problematic, and I hope to fix that soon. But apart from that, is there a problem e.g. with Equations?

Do we have a problem with unstructured/problematic/unmaintainable contributions? (We have literally 0 external contributors.) To me, the only contribution guideline we should ever have is "Please contribute your results to our library, we are extremely glad about every contribution from everybody."

mrhaandi:
In Dortmund I had now several internal talks about the library and got asked maintainability and software engineering related questions.
We currently little by little are ironing out potential maintainability hurdles (as does this PR, which addresses points I personally had trouble in the past). I like the wiki change pointing out "Other suggestions" as suggestions (not mandatory). So we do not have a maintainability problem with the current state of the library, and should welcome contributions not strictly abiding all the design guidelines. We currently have (on @dominik-kirst todo list) some FOL business which grew organically.
The other slight "offender" is `MuRec` which I had hard time understanding the somewhat scattered definitions (and the `_andrej` files).

DmxLarchey:
Dear @yforster and @mrhaandi,

This is not exactly on spot but I would like to react to the dependency issues that mutate very easily into _access issues_. We do not install the library so often so we may not pay attention to those issues but they are _real_ and they will almost certainly discourage trials by outsiders, or give reviewers a very bad first impression: they do not want to spend one day figuring out what is wrong with dependencies. And this is not only theoretical, we did witness this in several reviews for different papers.

Right now, __all__ _the installation procedures_ documented on the `coq-8.15` _default_ branch _fail_.

Several issues show up that I was able to find _only_ by digging into the `CI` configuration file `.github/workflows/build.yml`:

- `coq-library-undecidability.1.0.1+8.15` was not released, only `1.0.0+8.12` shows up. This is _pointed out_, but it does not help at solving the issue. If not available for `8.15` (but the same hold for `8.13+`), maybe we should not be put the _release install method_ upfront;
- `coq-8.15+rc1` is required but this is only available after `opam repo add coq-core-dev https://coq.inria.fr/opam/core-dev` is installed, which is not documented. Again this comes from the `CI` file;
- the `opam pin add` commands are not up to date so that `opam install . --deps-only` is unable to resolve dependencies. And the error message will not help in any way. Digging through the issue, I found that `opam pin add -n -y -k git coq-metacoq-template.dev+8.15 "https://github.com/MetaCoq/metacoq.git#3d83286"` was missing and is critical.

So yes, dependencies are a major cause of concern for a smooth access to the library for external users. I have the impression that the _default_ version of the library depend on non-stable libraries, or even non-stable Coq versions. 

It is good to have the `CI` but if it is not in sync with the documentation and contains some therefore hidden hacks, _it fails to check that the documented procedures work_.

DmxLarchey:
> The other slight "offender" is `MuRec` which I had hard time understanding the somewhat scattered definitions (and the `_andrej` files).

Hi @mrhaandi, what was your issue with `MuRec`? The dependencies with the `vec` and `pos` libraries? Notice that `vec` is now synonymous to `Vector.t` and `pos` to `Fin.y`. We could downgrade to non-dependent representation for µ-recursive algos but then, it would not be that natural and many edge-cases would show up in proofs.

mrhaandi:
> Hi @mrhaandi, what was your issue with `MuRec`? The dependencies with the `vec` and `pos` libraries? Notice that `vec` is now synonymous to `Vector.t` and `pos` to `Fin.y`. We could downgrade to non-dependent representation for µ-recursive algos but then, it would not be that natural and many edge-cases would show up in proofs.

About `MuRec` (`vec` etc. dependencies are fine)
1) Many files are on top-level (meaning all of them are most important?)
2) To understand `RA_UNIV_HALT.v` one needs to read through `recalg`, `rauniv`, `rauniv_andrej` (400 lines each). This goes against minimal, self-contained definitions. Minimal, self-contained definitions are handy for clarity and maintenance. For example, if some proof in `rauniv_andrej` breaks the definition in `RA_UNIV_HALT.v` is not compiled. 
3) I understand why some files are called `_andrej`, something more principled would be clearer. Or, why the more compilated construction still exists (as it is more cumbersome and provides no additional insight).

yforster:
> Dear @yforster and @mrhaandi,
> 
> This is not exactly on spot but I would like to react to the dependency issues that mutate very easily into _access issues_. 

Thank you for the summary. As I mentioned, we currently have the issue that MetaCoq has no released up-to-date `opam` packages. 

Just so I understand correctly: What we are currently missing is just the `opam repo add coq-core-dev https://coq.inria.fr/opam/core-dev` command? As you say, `opam pin add -n -y -k git coq-metacoq-template.dev+8.15 "https://github.com/MetaCoq/metacoq.git#3d83286"` is critical, but this is mentioned in the instructions.

DmxLarchey:
@yforster 

My mistake, in fact this `metacoq`

```
opam pin add -n -y -k git coq-metacoq-template.dev+8.15 "https://github.com/MetaCoq/metacoq.git#3d83286"
```

does _not_ work, it should be replaced with `template-coq`

```
opam pin add -n -y -k git coq-metacoq-template.dev+8.15 "https://github.com/yforster/template-coq.git#dad61c22c3"
```

which I got from the `CI` file. Is this your personal fork of Metacoq ?

DmxLarchey:
The structure in `MuRec` has to be improved I agree. I will put this on my TODO list but I am busy with other things right now.

> 2. To understand `RA_UNIV_HALT.v` one needs to read through `recalg`, `rauniv`, `rauniv_andrej` (400 lines each). This goes against minimal, self-contained definitions. Minimal, self-contained definitions are handy for clarity and maintenance. For example, if some proof in `rauniv_andrej` breaks the definition in `RA_UNIV_HALT.v` is not compiled.

Ok I see. I do not think there is solution to that: there is no easy and self-contained way to describe a universal machine in extension, at least that I know of. Their are indeed two with measured size is around 10k each. One of the µ-rec `ra_univ` solves elementary Diophantine constraints, the other one `ra_univ_andrej` solve the uniformized version (possibly shorter that you proposed). The given file `RA_UNIV_HALT.v` has to load these algorithms one way or another, because the problem to solve is: does this _fixed_ particular µ-rec halt on a given natural number as input. 

> 3. I understand why some files are called `_andrej`, something more principled would be clearer. Or, why the more compilated construction still exists (as it is more cumbersome and provides no additional insight).

Indeed we could clear out or deprecate one of the constructions into a sub-dir. There is no point having two except for those who would like to study how it is precisely implemented. Add to the TODO as well.

PR #140, opened 2022-02-07
Typo in the definition file PCP.v impacts naming schemes

DmxLarchey:
This is just an example of how correcting a typo impacts automatic naming schemes possibly far away in the code.

In `PCP.v`, cards are pairs of `string X` because `list` are understood as finite sets there. Of course they are synonymous but since `string` was defined, why not use it?

Well, this impacts introduced names where the default `l` is now replaced with `s`.

There is no need for this PR to be merged except if you agree that the typo should be corrected


mrhaandi:
Actually, why `Definition string X := list X.` instead of `Notation`? Sometimes layers of definitions introduce difficulties.

mrhaandi:
Also why `coq.8.15+rc1`? I'm currently working with the main `coq.8.15` release.

DmxLarchey:
> Also why `coq.8.15+rc1`? I'm currently working with the main `coq.8.15` release.

Well this is the version in the current [`opam`](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.15/opam) file on the `coq-8.15` branch. I did change it for a test and forgot to revert it back.

DmxLarchey:
> Actually, why `Definition string X := list X.` instead of `Notation`? Sometimes layers of definitions introduce difficulties.

I do not know. I did not change that. Indeed `Notation` could be better. What do you think @yforster?

yforster:
> Also why `coq.8.15+rc1`? I'm currently working with the main `coq.8.15` release.

I have filed a PR updating to `coq-8.15` and fixed the installation instruction which were broken as discovered by Dominique.

> Actually, why Definition string X := list X. instead of Notation? Sometimes layers of definitions introduce difficulties.

Actually, this here immediately introduces a universe problem. So yes, both string and card should be a `Notation`

DmxLarchey:
Ok I fix this as a `Notation`, then I wait for the updated `coq-8.15` and propose a merge when this works. I may have to revert back some names. I will try to remove the use of auto-generated names, if there is a quick fix.

DmxLarchey:
@mrhaandi @yforster. Rebase done. I will merge tomorrow if no one objects.

PR #143, opened 2022-02-10
Simple Binary Turing Machines (2) RFC

mrhaandi:
This PR is a request for comments on a minimal TM formulation.

The current [SBTM.v](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.15/theories/TM/SBTM.v) has some drawbacks as a minimal Turing machine.

Most crucial points are:
1) The tape `([true], None, [true])` is a valid input for `HaltSBTM`, but it has no semantics. The head is not pointing at a cell (or tape end). It also has no interpretation in `TM.tape`.
2) The transition function `trans` has a ternary alphabet `Some true; Some false; None` (or, let's say 2.5-ary since `None` has special meaning, but can be used for computation)
3) The had can move in three ways (move left, move right, do not move)

Points 2. and 3. create 9 cases in proofs, which I wish to minimize.
Point 1. can really interfere with reductions because one needs to consider/encode ill-formed tapes. At one point I had to use `TM bool 1` instead of `SBTM` because of this.

The proposed PR contains a Turing machine formalization (currently called `SBTM2`) with the following properties:
1) The only two symbols are blank (`false`) and non-blank (`true`), so it is truly binary.
2) The tape is infinite in both directions and implicitly is filled with blanks.
3) At each operation the head moves either left or right. 
4) There are no special conditions for the end of tape, because the tape has no end.
5) The execution is defined via a (deterministic) step function.

Points 1. and 3. create only 4 cases in proofs.
I intend to rename `SBTM2` into something more proper. Besides your comments, please suggest names.

I also intend to have `SBTM2_HALT0` as a seed, which is halting on the tape filled with blanks.


theories/TM/SBTM2.v
```diff
@@ -0,0 +1,58 @@
...
+  | go_right =>
+      match t with
+      | (ls, a, r :: rs) => (a :: ls, r, rs)
+      | (ls, a, nil) => (a :: ls, false, nil)
+      end
+  end.
+
+Record SBTM2 := {
+  num_states : nat;
+  trans : Fin.t (S num_states) * bool -> option (Fin.t (S num_states) * bool * direction) }.
```
DmxLarchey, line 35: Is there a reason for having at least one state and carrying this `S num_states` around ? I know it does not make much sense to having no state at all but does it harm to include that trivial case?

DmxLarchey:
Nice compact description of TM with finitary tapes indeed. Rather than infinite, I would qualify this as _unbounded_, or _automatically expands when needed_. But it can only contain a finite amount of information. It is isomorphic to `{ f : Z -> bool | kuratowki_finite (fun z => f z = true) }` where `kuratowki_finite` is an inductive characterization of _listable_, ie closed under singletons, empty and binary union.

Could indeed be used as a seed, but do you have a specific use case in mind?

DmxLarchey:
> Do you see a chance to have an easy reduction `HaltSBTM2 <= Halt_TM`

or perhaps `HaltSBTM2 <= BSM` ?

yforster:
> > Do you see a chance to have an easy reduction `HaltSBTM2 <= Halt_TM` (i.e. a multi-tape Turing machine simulating SBTM2s)? When we were thinking about replacements for `HaltL` as ancestral seed in the def of undecidability we though about using `SBTM`, but that is unsatisyfing because it's not clear ho to close the loop with not-so-simple Turing machines.
> 
> I would say yes, going to `Halt_TM` even in the boolean, single tape should be easy. Identify `None` with `false` and write `false` if one reaches the end of the tape.

Ha, I see.

The simulation relation might be a little tricky though, right? If an SBTM2 steps right on `([], true, [])` we end up with `[true], false, []`. A single-tape TM simulation of this would start at `midtape [] true []`, but then either go to `righttape [true]`, or would have to check after every step whether the end of the tape is reached, and if so fix it to get back into `midtape [true] false []`.

mrhaandi:
> The simulation relation might be a little tricky though, right? If an SBTM2 steps right on `([], true, [])` we end up with `[true], false, []`. A single-tape TM simulation of this would start at `midtape [] true []`, but then either go to `righttape [true]`, or would have to check after every step whether the end of the tape is reached, and if so fix it to get back into `midtape [true] false []`.

I did write a tape equivalence relation for such a problem (when just walking around tape ends introduces redundant symbols).
So this I do not think this will become an issue (simulation will be up to tape equivalence).

theories/TM/SBTM2.v
DmxLarchey: Ok I get that this gives you a default starting state. Assuming `s₀ : Fin.t n` would be equally good? I mean, I don't see why you would use the fact that `Fin.F1` is the least one? I guess you just use `Fin.t n` because it is finite. Having `s₀` preserves the symmetry in definition. But this is a matter of taste of course.

I agree that the dependent match is to be avoided.

mrhaandi: `s₀ : Fin.t n` would need to be part of the problem instance.
Sometimes starting at state zero is (or the empty tape, which I will include) is much more convenient.

I still have a consideration: if `n = 0`, then the machine has no states making a traditional configuration notion ill-formed.

DmxLarchey:
> > It is isomorphic to `{ f : Z -> bool | kuratowki_finite (fun z => f z = true) }` where `kuratowki_finite` is an inductive characterization of _listable_, ie closed under singletons, empty and binary union.
> 
> Well, isomorphic under funext + PI, or isomorphic with a handmade def of isomorphism just for this case.

Sure I just meant what you can do with one, you can do with the other, w/o funext+PI. I should have written "isomorphic".

> So `{f : Z -> bool | exists l : Z, forall z, f z = true <-> In z l}`? No reason to pick a weaker notion of finiteness than... well, actual finiteness.

Actually I was happy to discover, thanks to Martin Escardo in a twitter exchange, that _listability_ (which I have been using as a definition for finiteness for a very long time) was already characterized as Kuratowski finite, ie they are equivalent. Provided of course you have access to the list inductive datatype. You can however define Kuratowski has the least predicate closed for the above rules (which does no require inductive predicates) and can be done in the calculus of construction alone, or as it comes from there, set theory.

theories/TM/SBTM2.v
mrhaandi: I followed the suggestion and changed the definitions to
`trans : (Fin.t num_states) * bool -> option ((Fin.t num_states) * bool * direction)` and
```
Definition SBTM2_HALT : { M : SBTM2 & config M } -> Prop :=
  fun '(existT _ M x) => exists k, steps M k x = None.
```
This made the reduction slightly more compact and I hope will make reductions from also more uniform.
The downside is that if `num_states = 0`, then the configuration space is empty. Hopefully, this is not a big deal.
I'll see whether for potential `HaltSBTM2 <= Halt_TM` this maks a difference.

DmxLarchey:
> Do you think things would become much harder by going to an association list instead of a finite function @mrhaandi?

or possibly ?

```
Record SBTM2 := Build_SBTM2 {
  num_states : nat;
  trans_true : vec (option ...) num_states;
  trans_false: vec (option ...) num_states;
}
```

yforster:
Alternative:

```
Record SBTM2 := Build_SBTM2 {
  num_states : nat;
  trans : vec (option ... * option ...) num_states
```

Or with a starting state:

```
Record SBTM2 := Build_SBTM2 {
  num_states : nat;
  start : Fin.t num_states ;
  trans : vec (option ... * option ...) num_states
```

(why did I not use this for `SBTM`?)

mrhaandi:
I assumed the finite function is by design to be easy to understand. Going to a vector of pairs is not too much work.
I stayed close to the old `SBTM` to not have something completely arbitrary looking.

yforster:
Yes, it was by design. But that ruled it out for a seed, which didn't seem to be a problem for me. But now we are thinking about a better ancestral seed, so it might be one way to go.

mrhaandi:
As suggested, I have now defined `SBTM2` via a transition table rather than finite function:
```
Record SBTM2 := Build_SBTM2 {
  num_states : nat;
  (* transition table *)
  trans : Vector.t (
    (option ((Fin.t num_states) * bool * direction)) *
    (option ((Fin.t num_states) * bool * direction)))
    num_states }.
```
The `step` function is implemented via table lookup. And there are lemmas to switch betwen table/function representations, so the proofs are mostly unchanged.

mrhaandi:
@yforster From what I understand, the necessary reduction to implement is `SBTM2 < SRH` (from which `SBTM2 < BSM` follows vis `PCP`). The direct `SBTM2 < BSM` is a shortcut which would be nice to have but not strictly mandatory, right?
In this case, I would implement `SBTM2 < SRH` and remove `SBTM` (renaming `SBTM2` to `SBTM`).
As for `BSM`, I'm not that familiar with the underlying framework, so could you take a look into the shortcut? I expect the reduction to be actually simpler than from the old 9-case `SBTM` (and the default `false` should make things uniform here too).

mrhaandi:
I have now added the reduction `SBTM2_HALT <= SRH`. (Old names and chains with `SBTM` are still in).
Comparing the old [HaltSBTMu_to_SRH.v](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.15/theories/StringRewriting/Reductions/HaltSBTMu_to_SRH.v) with the new [SBTM2_HALT_to_SRH.v](https://github.com/mrhaandi/coq-library-undecidability/blob/sbtm2/theories/StringRewriting/Reductions/SBTM2_HALT_to_SRH.v) there are a few things immediately visible:
- before there are 18 rules for encoding the transition, now there are just 5 (or 7 depending on counting)
- all proofs are by direct case analysis without custom (quite scary) Ltac
- still the overall spelled out proof is not longer
- the new proof is arguably more maintainable

mrhaandi:
Since the reductions `SBTM2_HALT <= SR` and `SBTM2_HALT <= TSR` (now also included) are simple, I removed the intermediate problem `SRH` from reductions chains.
Now we directly have `HaltTM_1 <= SBTM2_HALT <= SR <= ...` instead of the previous `HaltTM_1 <= HaltSBTM <= HaltSBTMu <= SRH <= SR <= ...`

DmxLarchey:
> As for `BSM`, I'm not that familiar with the underlying framework, so could you take a look into the shortcut?

I will give it a try. Can I push to this PR?

mrhaandi:
> > As for `BSM`, I'm not that familiar with the underlying framework, so could you take a look into the shortcut?
> 
> I will give it a try. Can I push to this PR?

I'm currently also out of curiosity looking into this. Maybe I'll be done in a few hours.

mrhaandi:
> > As for `BSM`, I'm not that familiar with the underlying framework, so could you take a look into the shortcut?
> 
> I will give it a try. Can I push to this PR?

Yes, the proof `SBTM2_HALT <= HaltBSM` (now checked in) is quite easy since the program is just repeating 13 instructions.
I ran only into the following complications: sometimes `subcode_tac` got confused and I had to manually write 
```
{ eexists [_; _; _; _; _; _; _; _; _; _; _], _; by split; [|cbn; lia]. }
```
I did not find a lemma (and wrote my own) stating something like
```
out_code i (1, P) -> (1, P) // x -[m]-> y ->  (1, P) // x -[n]-> (i, v) -> (1, P) // y -[n - m]-> (i, v).
```
which is oftentimes useful for machine simulations.

DmxLarchey:
> I did not find a lemma (and wrote my own) stating something like
>
> ``` 
> out_code i (1, P) -> (1, P) // x -[m]-> y ->  (1, P) // x -[n]-> (i, v) -> (1, P) // y -[n - m]-> (i, v).
> ```
> which is oftentimes useful for machine simulations.

First that one. Something a bit more general exists in [`Shared/.../Code/sss.v`](https://github.com/uds-psl/coq-library-undecidability/tree/coq-8.15/theories/Shared/Libs/DLW/Code/sss.v#L593) which is 

```
subcode_sss_subcode_inv :
            out_code (fst st3) P 
        -> P <sc Q
        -> P // st1 -[p]-> st2
        -> Q // st1 -[k]-> st3
        -> exists q, k = p+q /\ Q // st2 -[q]-> st3.
```

It works with subcode as well. I agree this is indeed a very useful lemma. Beware that it works only for _deterministic models_ of computation, which is the case for `BSM`.

DmxLarchey:
> I ran only into the following complications: sometimes `subcode_tac` got confused and I had to manually write

Yes that tactic tries to check that some `subcode` occurs at the right position in bigger code. It is kind of a hack and there are indeed pitfalls. Typically, if you have two "identical" occurrences of the same `subcode`, it will not try the second one. The work around I use in general is the following

```
Let i' := i.

Definition code := .... :: POP x i j :: .... :: POP x i' j :: ... .
```

Then `subcode_tac` will distinguish `i` from `i'` and find the proper occurrence. I don't know if this was your issue but for me, duplicated occurrences is almost always the reason why it fails.

DmxLarchey:
Btw nice compiler in a single step. It is probably gonna be shorter than the one I am currently implementing but it seems to me you use 4 stacks. I  started with 2 (which creates a disymmetry) because one of the stacks has to store that value under the head and I am currently dealing with this issue. 

I get why you would use 3 stacks, to store the value under the head on a separate stack. But what use is the fourth stack used for ?

Also, clearly `BSM₂` with 2 stacks is undecidable because `MM₂` is. But what do you think of `BSM₁` with one stack only?

mrhaandi:
> `subcode_sss_subcode_inv`

Thank you for the pointer, indeed this is what I was looking for.

mrhaandi:
> Btw nice compiler in a single step. It is probably gonna be shorter than the one I am currently implementing but it seems to me you use 4 stacks. I started with 2 (which creates a disymmetry) because one of the stacks has to store that value under the head and I am currently dealing with this issue.
> 
> I get why you would use 3 stacks, to store the value under the head on a separate stack. But what use is the fourth stack used for ?
> 
> Also, clearly `BSM₂` with 2 stacks is undecidable because `MM₂` is. But what do you think of `BSM₁` with one stack only?

I just copied the previous `SBTM` approach. The three stacks are: left tape, current symbol, right tape, always nil.
The always nil stack can be used to have an unconditional jump within one instruction.
I considered using only two stacks, but the construction would become larger.
`BSM_1` should be your usual pushdown autotomaton with decidable termination, right?

theories/StackMachines/Reductions/SBTM2_HALT_to_HaltBSM.v
```diff
@@ -0,0 +1,310 @@
...
+  #[local] Arguments encode_state : simpl never.
+  #[local] Notation "! p" := (encode_state p) (at level 1).
+
+  Definition encode_config '(q, t) : bsm_state 4 := (!q, encode_tape t).
+
+  #[local] Notation JMP i := (POP ZERO i i).
+  
+  Definition box '(q, a) (f : (state M * bool * direction) -> bsm_instr 4) : bsm_instr 4 :=
+    match δ (q, a) with
+    | None => JMP 0
```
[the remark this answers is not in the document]
mrhaandi: Yes, the need for two instructions is why I used the empty fourth stack (as in the previous construction), which is fine for the goal.

mrhaandi:
> Let i' := i.

This hack may break at some point. I followed your suggestion and have hidden a duplicate behind a seperate `Definition`. Now the tactic is able to solve it.

I also now use `subcode_sss_subcode_inv` and have removed my own lemma, which shortens the proof.

mrhaandi:
I wonder whether this shorter path (without `PCP`) should be used e.g. for `MinskyMachines/Reductions/HaltTM_1_to_MM`?

DmxLarchey:
> I wonder whether this shorter path (without `PCP`) should be used e.g. for `MinskyMachines/Reductions/HaltTM_1_to_MM`?

I think that is a good idea. Moreover this path could show that we can simulate not only termination but also output values, as in the difference between `H10` and `DPRM`. I am not sure 100% but passing through decision problems like `PCP` kind of prevents that.

DmxLarchey:
> @DmxLarchey there was a bug (it did complain about missing variables) with `bsm inv POP empty`, which now should be fixed; please check.

Indeed I guess I wrote a tactic I never actually use!! Otherwise the bug should have popped up.

yforster:
> > I wonder whether this shorter path (without `PCP`) should be used e.g. for `MinskyMachines/Reductions/HaltTM_1_to_MM`?
> 
> I think that is a good idea. Moreover this path could show that we can simulate not only termination but also output values, as in the difference between `H10` and `DPRM`. I am not sure 100% but passing through decision problems like `PCP` kind of prevents that.

All of the input/output simulations are done, but not merged. They are here: https://github.com/yforster/coq-library-undecidability/tree/forthethesis/

DmxLarchey:
>  I considered using only two stacks, but the construction would become larger. 

Yes indeed !!

> `BSM_1` should be your usual pushdown automaton with decidable termination, right?

Do you think this one could be included in the decidability results section of the library? I mean is it reasonably doable?

DmxLarchey:
> All of the input/output simulations are done, but not merged. They are here: https://github.com/yforster/coq-library-undecidability/tree/forthethesis/

The code in [TM_computable_to_BSM_computable.v](https://github.com/yforster/coq-library-undecidability/blob/forthethesis/theories/StackMachines/BSM/TM_computable_to_BSM_computable.v) is ... complicated ;-) do you have a reason for hiding machine code inside proofs (ie using Sigma types) ?

mrhaandi:
As previously discussed, I renamed `SBTM2` into `SBTM` and replaced the previous reductions by the new ones.

mrhaandi:
@DmxLarchey The somewhat hacky `bsm sss ...` structure has the drawback that every step (although it is given by specification) also needs to be explicitly recalled in the proof (for every case).
In the recent commit I redesigned the proof of `PROG_spec_Some` to be mostly by computation. This has several benefits:
1) the proof length is reduced from 70 lines to just 6 lines because one does not need to recall every instruction, and `reflexivity` does almost everything
2) the compilation time is reduced a little because there are fewer constraints
3) no custom ltac needed (such as `subcode_tac `), just lemma application suffices
4) for free the `PROG_spec_Some` lemma is strengthened to `-[S k]->` instead of `-[k]->` which simplifies `inverse_simulation`

The only drawback is that currently `Q_step` and `Q_step_spec` which are general lemmas are written ad-hoc. Is there something for this in your framework?

mrhaandi:
> Ok so as I understand it, `Q_step` realizes/computes the small-step semantics within Coq. Could `Q_step` also work for programs not starting at `1` but at say `10+i` ?

`Q_step` has `offset` (global position) and `i` (local position within subprogram) as arguments. So the answer is yes and no.
Yes in the sense that it does not matter at what `offset` the subprogram is.
No in the sense that computation has to occur locally within the subprogram where `i` addresses starting from `0`.
It is quite handy to verify subprograms automatically by computation which can be seen in `PROG_spec_Some`.
Of course, it can only reason locally. Still this is the usual case (and effective for the current reduction) that the whole program consists of simple building blocks.

mrhaandi:
As @yforster pointed out, we may consider making the new `SBTM_HALT` the initial undecidability seed.
For this I added all necessary enumerators, and we have 
1) enumerable `{ M : SBTM & config M }` (as problem space)
2) enumerable `SBTM_HALT`
3) decidable (`SBTM_HALT`) -> enumerable (complement `SBTM_HALT`).
4) decidable (complement `SBTM_HALT`) -> enumerable (complement `SBTM_HALT`). (needed for reductions from complement)

What is your opinion on the current state? Shell we switch to `SBTM_HALT`? Is has all the nice properties and enjoys direct reductions from/to `HaltTM 1`.

yforster:
The error message in CI is just due to `Set Implicit Arguments`. Probably a `Global Set Implicit Arguments` leaks via `L` into this file, and now is not there. The easy fix is to just add an underscore at the position.

mrhaandi:
> Sorry made a click mistake ... I solved the issue in `BPCP_SigBPCP.v` ... indeed an problem with implicit/explicit arguments. Concerning `TRAKHTENBROT_to_FSAT.v`, mainly `now intuition` failed (that's @dominik-kirst 's, I do not use `intuition` ;-)
> 
> Also strangely, an equation was reverse by `symmetry` ?

`cbn; try now intuition; try firstorder.` with several goals left. I see how this code breaks on the slightest touch.
Thanks @DmxLarchey for pushing the fix. (Also, I previously removed the `ssrfun.obind` reference as suggested).

yforster:
> Also strangely, an equation was reverse by `symmetry` ?

I have encountered that issue multiple times while porting between Coq versions. No idea where it's coming from, potentially something internal to `firstorder` / `intuition`?

mrhaandi:
I updated the readme, stating the new definition of undecidability via `SBTM_HALT`. So the overall PR looks ready for review and merge.

DmxLarchey:
@yforster and @mrhaandi, I encountered troubles several times when `*.v` files are moved around between branches: after switching with `git checkout`, there may exist `*.vo` files than cannot be cleaned up by `make clean` anymore because they do not show up with the correct path in `_CoqProject`. So I am proposing a `make superclean` target to deal with this problem easily.

mrhaandi:
> @yforster and @mrhaandi, I encountered troubles several times when `*.v` files are moved around between branches: after switching with `git checkout`, there may exist `*.vo` files than cannot be cleaned up by `make clean` anymore because they do not show up with the correct path in `_CoqProject`. So I am proposing a `make superclean` target to deal with this problem easily.

I prefer `git clean -fxd` to have a clean directory (the drawback, of course, is that it may remove important uncommitted files).
There are also other `git` commands to remove only files that would be ignored anyway (such as `.vo`).

yforster:
When switching between branches, everything usually becomes recompiled anyways. So a `git clean -xfd` (equivalently to Andrej's suggestion, just memorised differently in my muscle memory) doesn't do any harm.

What would your proposed `make` target be?

I am sometimes also using Fabian's shell scripts to just compile one file and all its dependencies and then remove all other files which are not relevant: https://gist.github.com/fakusb/0e1c6256fd8344a65d7cf995dfdca162

DmxLarchey:
> I prefer `git clean -fxd` to have a clean directory (the drawback, of course, is that it may remove important uncommitted files). There are also other `git clean` flags such as `-X` to remove only files that would be ignored anyway (such as `.vo`).

Thanks that is an alternate solution of course. But does it harm if something become a `make` target ?

Btw, what do you think of a target making a recursive `coqwc` on the whole library @yforster and @mrhaandi ?

mrhaandi:
> I have a tendency to keep `Makefile`s simple and just use the relevant command when I want to use them. But maybe it's not harmful to have them as well.

I also prefer to have only the essentials in the makefiles (also because I'm using Windows/wsl/cygwin/mingw where some things just do not exist or miserably fail).

DmxLarchey:
The `coqwc` target would be an extra maybe.

But I think a `superclean` or `mrproper` target would be helpful. The error message you get in case of hanging out `.vo` files masking/duplicating legitimate ones is not that clear for a newcomer.

DmxLarchey:
> But I think a `superclean` or `mrproper` target would be helpful. 

Regardless of the actual implementation with `find` or `git clean` ...

yforster:
Everything looks great, I'm just unsure about one thing:

I personally really dislike the presentation of any calculus with a step function and then defining evaluation using iteration. To me, those are internal things to a functionality proof. Evaluation should be an inductive proposition, because that's a lot easier to understand and doesn't require weird definitions like `obind` (not saying that's weird in general, but it's very weird to have it in a *definition*).

The definitions of `L` and `TM` are already in this fashion. In the above mentioned branch, I also have ported all other halting problems of our main models of computation to an `eval` predicate (of course keeping the old notions via equivalence lemmas). I want to merge that at some point, so maybe now is a good point to discuss this issue.

mrhaandi:
If I understand @yforster correctly, the main question is what to put as `?` into 
`Definition undecidable {X} (p : X -> Prop) := decidable p -> enumerable (complement ?)`
the functional/ad-hoc inductive/generic inductive formulation?
Currently, we are three people with three opinions based on flavor.

However, creating `SBTM_equiv.v` in addition to `SBTM.v` currently goes a little against design principles. Maybe we can also revisit those. Lets not open too many cans at once. So we can put all three formulations into modules in `SBTM.v`.

My suggestion is leave the current state as is, and as the next step since @yforster intends to merge a uniformization of different computational models, add such an ad-hoc inductive evaluation formulation (to any relevant model).

mrhaandi:
@yforster @DmxLarchey 
Shall we then merge the current state and continue from there?

DmxLarchey:
I did agree with the initial merge and the PR has improved since then. I agree more.

DmxLarchey:
Hi @yforster. Do you object in merging this PR? In case not, I will merge it next week.

yforster:
I'm on holidays and won't object anything :)

DmxLarchey:
> I'm on holidays and won't object anything :)

Ok I will merge then. Enjoys your holidays ;-)

PR #145, opened 2022-03-11
HaltL <= MMA_HALTING (natural and fast)

mrhaandi:
This PR is adds two well-known problems (and their variants for closed terms):
- weak call-by-name leftmost outermost normalization for lambda-terms `wCBN`
- Krivine machine halting `KrivineM_HALT` [1, 2]

These natural intermediate steps allows for an easy reduction from L halting `HaltL` to (alternative) Minsky machine halting `MMA_HALTING`. This is the main contribution of this PR, and establishes a quick (almost stand-alone) connection from `L` to `MMA`. In particular, we get `HaltL <= wCBNclosed <= KrivineMclosed_HALT <= MMA_HALTING`.

Additionally, the reduction `MMA_HALTING n <= HaltTM n` closes the loop to Turing machines.

Remarks:
- Overall compilation time of all components (on my machine) is under 10 seconds.
- Overall code size is 2000 LOC (mostly no custom LTAC or `mma` framework use).
- wCBV simulation by wCBN is not done by a self-interpreter, but by Plotkin's continuation translation 
- Krivine machine halting `halt_cbn` is entirely inductive and easy to simulate by `mma`.
- I did not use `2^n * 3^m` as pairing for `mma`, but instead `(n+n+1) * 2^m` which is easier to implement and does suffice.
- Capturing de Bruijn substitutions in `HaltL` do not behave well, and require effort to work with. For example, it does not suffice to replace free variables by ((\x.xx) (\x.xx)) to get from open to closed non-termination. I would go as far as saying that `HaltL` is ill-defined for open terms because it has no correspondence in the traditional, named setting. (This is an unfortunate design decision.)
- Currently, `HaltLclosed` is used as an intermediate step with a lot of machinery, which is entirely due to arbitrary behavior of capturing de Bruijn substitutions. Traditionally, this would require almost no effort.

Potential TODOs:
- [x] Make `HaltLclosed` more prominent and move it away from Turing machines.
- [x] @DmxLarchey is there a MMA_HALTING <= MMA2_HALTING reduction?
- I see little point in MM_HALTING (where the jump is on decrease 0). However, there is a lot of code for it, which to me looks obsolete.
- [x] Update README.

[1] Krivine, Jean-Louis. "A call-by-name lambda-calculus machine." Higher-order and symbolic computation 20.3 (2007): 199-207.
[2] https://en.wikipedia.org/wiki/Krivine_machine


yforster:
> Currently, `HaltLclosed` is used as an intermediate step with a lot of machinery, which is entirely due to arbitrary behavior of capturing de Bruijn substitutions. Traditionally, this would require almost no effort.

How would you prove that halting for open terms w.r.t. *weak* cbv/cbn reduces to halting for closed terms "traditionally"?

DmxLarchey:
`MM_HALTING` could be replaced completely with `MMA_HALTING` but then we would need to rework `BSM_to_MM_HALTING` and `MM_HALTING` to `FRACTRAN` and the papers which refer to that code wouldn't be accurate anymore (CPP'19 and FSCD'19 at least).

I would agree that we could have chosen `MMA_HALTING` over `MM_HALTING` from the start but the issue with 2 counters popped later on and the "damage" was done already. My mistake may be. I did not realize the zero/not zero jump was an issue before.

theories/Shared/Libs/DLW/Vec/vec.v
```diff
@@ -162,6 +162,20 @@ Section vector.
     + repeat rewrite vec_change_neq; auto.
   Qed.
 
+  Lemma vec_change_comm n v p q x y : p <> q ->
+    vec_change (@vec_change n v p x) q y = vec_change (vec_change v q y) p x.
+  Proof.
+    intros Hpq.
+    apply vec_pos_ext; intros r.
+    destruct (pos_eq_dec p r) as [?|Hpr].
+    + subst. rewrite (vec_change_neq _ _ (not_eq_sym Hpq)).
```
DmxLarchey, line 171: May be the `rew vec` tactic solves this by itself preceeded with `destruct (eq_nat_dec r p)` and `destruct (eq_nat_dec r q)` ?

theories/Shared/Libs/DLW/Code/sss.v
```diff
@@ -702,6 +677,38 @@ Section Small_Step_Semantics.
...
+  Proof.
+    intros (st2 & (k & H1) & H2); revert st st2 H1 H2.
+    induction k as [ k IH ] using (well_founded_induction_type lt_wf).
+    intros st1 st2 H1 H2.
+    destruct k as [ | k ].
+    + apply sss_steps_0_inv in H1.
+      subst; apply HR0; auto.
+    + apply HR1.
+      intros Q st2' HQ H5.
+      assert (P // st2' ↓) as (st3 & H6 & H7).
```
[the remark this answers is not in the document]
DmxLarchey: Ok I did not see that ` P // st2 ↓` was removed in the induction hypothesis. So that is fine with me. Didn't this change impact existing proofs? Not so many would use it but still.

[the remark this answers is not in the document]
mrhaandi: I was trying to use `sss_terminates_ind` but it was too cumbersome (and also was not used in the library before?).
The issue was the additional proof obligation `P // st2 ↓` in previous version of `HR1`.
It can be proven locally, hence the assert `assert (P // st2' ↓)`. Now `sss_terminates_ind` is easier to use and I replaced my previously functional proofs in with it. This removed a good amount of auxiliary code.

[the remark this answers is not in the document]
DmxLarchey: Ok thx

theories/Shared/Libs/DLW/Vec/vec.v
mrhaandi: Yes, I now modified the proof to use `rew vec`.

mrhaandi:
> When compressing to 2 registers, there is no way out of the Gödel coding. One can do otherwise (simpler ?) with more than 3 registers, compressing along e.g. the `(2b+1).2^a` encoding.

Yes, as discussed yesterday this step is sensible only with Gödel coding.
The suggested compression `3+n` to `2+n` would replace first (or last?) three counters `a`, `b`, `c` by two counters: one auxiliary, and one holding `2^a * 3^b * 5^c`; all remaining counters would not be touched.
By "simpler" I did not mean that Gödel coding is avoided or that the resulting machine is small. Rather, only multiplication/division by `2, 3, 5` is required (no machinery for arbitrary primes). Also one does not need first to compress to three and then to two counters using different techniques. The resulting machine for iterated compression (starting from five counters) would be larger and less efficient. Still, I believe that this would result in the smallest/fastest proof script overall.

DmxLarchey:
> one motivation for the simplest proof script is the extension of the compression technique to reversible counter machines

OK I see better now. I will give a try, hopefully soon.

mrhaandi:
@DmxLarchey A side remark regarding `Theorem sss_terminates_ind st : P // st ↓ -> R st.` (and similar).
The Coq standard library contains `Lemma Acc_clos_trans : forall x:A, Acc R x -> Acc trans_clos x.`
That means for a functional step relation we can:
First, show that if a configuration is terminating, then it is accessible wrt. the inverse step relation (easy).
Second, apply `Acc_clos_trans` to get accessibility for the transitive closure.
Induction on the resulting assumption (`Acc (clos_trans ...) st`) subsumes `sss_terminates_ind`.

Using this, I already removed some boilerplate code in my own proofs (no more juggling step counters). Maybe you find the idea useful.

DmxLarchey:
> First, show that if a configuration is terminating, then it is accessible wrt. the inverse step relation (easy).

Yep Termination and Strong Normalisation are equivalent for functional relations.

However, `P // st ↓` does not exactly means termination, but instead, reaches a instruction out of the code (which is equivalent to termination for total step relations).

mrhaandi:
I added the almost trivial `MMA_HALTING n <= HaltTM n` reduction to close the loop.

Now we have a fast, alternative (without the complexity framework?) path from `HaltL` to `HaltTM 1`:

```coq
Lemma HaltL_to_HaltTM_1 : HaltL ⪯ HaltTM 1.
Proof.
  apply (reduces_transitive HaltL_to_HaltLclosed.reduction).
  apply (reduces_transitive HaltLclosed_to_wCBNclosed.reduction).
  apply (reduces_transitive wCBNclosed_to_KrivineMclosed_HALT.reduction).
  apply (reduces_transitive KrivineMclosed_HALT_to_MMA_HALTING.reduction).
  apply (reduces_transitive MMA_HALTING_n_to_HaltTM_n.reduction).
  apply (reduces_transitive (mTM_to_TM.nTM_to_MTM 5)).
  exact mTM_to_TM.MTM_to_stM.
Qed.
```

yforster:
We might still depend on MetaCoq for the reduction to HaltLclosed. But I think we even have a hand-written self-interpreter for L somewhere thag we can use instead

mrhaandi:
> We might still depend on MetaCoq for the reduction to HaltLclosed. But I think we even have a hand-written self-interpreter for L somewhere thag we can use instead

The main issue is the ill-defined behavior of `HaltL` on open terms. I would argue that `HaltLclosed` is the actual halting problem at hand and `HaltL` an artifact.

mrhaandi:
@yforster @DmxLarchey 
I consider this PR good to go. I can merge as soon as you accept as reviewers.

README.md
```diff
@@ -78,6 +79,7 @@ Target problems are very expressive and thus work well as targets for reduction,
 - Uniform boundedness of deterministic, length-preserving stack machines (`SMNdl_UB` in [`StackMachines/SMN.v`](theories/StackMachines/SMN.v))
 - Semi-unification (`SemiU` in [`SemiUnification/SemiU.v`](theories/SemiUnification/SemiU.v))
 - System F Inhabitation (`SysF_INH` in [`SystemF/SysF.v`](theories/SystemF/SysF.v)), System F Typability (`SysF_TYP` in [`SystemF/SysF.v`](theories/SystemF/SysF.v)), System F Type Checking (`SysF_TC` in [`SystemF/SysF.v`](theories/SystemF/SysF.v))
+- Halting problem Krivine machines (`KrivineM_HALT` in [`LambdaCalculus/Krivine.v`](theories/LambdaCalculus/Krivine.v))
```
yforster, line 82:
```suggestion
- Halting problem for Krivine machines (`KrivineM_HALT` in [`LambdaCalculus/Krivine.v`](theories/LambdaCalculus/Krivine.v))
```

yforster (approved):

DmxLarchey (approved):

README.md
mrhaandi: Done.

PR #149, opened 2022-05-31
global Set Default Proof Using "Type"

mrhaandi:
This PR makes `Set Default Proof Using "Type"` uniform for the library (and future PRs). In particular:
- `Default Proof Using "Type"` is set globally via `_CoqProject`
- all individual `Set Default Proof Using "Type"` are removed
- additional, necessary `Proof using` are added
- `vos` compilation time slightly decreases from `user 9m37.746s` to `user 8m59.793s`
- `ReducibilityFacts` is refactored for minimal dependencies and more graceful failure

### DIFF
--- a/theories/H10/H10_undec.v
+++ b/theories/H10/H10_undec.v
@@ -14,6 +14,7 @@ From Undecidability.Shared.Libs.DLW
 
 From Undecidability.Synthetic Require Import Undecidability.
 Require Import Undecidability.Synthetic.ReducibilityFacts.
+Require Import Undecidability.Synthetic.MoreReducibilityFacts.
 
 Require Import Undecidability.TM.TM.
 
--- a/theories/H10/H10_undec.v
+++ b/theories/H10/H10_undec.v
@@ -14,7 +14,6 @@ From Undecidability.Shared.Libs.DLW
 
 From Undecidability.Synthetic Require Import Undecidability.
 Require Import Undecidability.Synthetic.ReducibilityFacts.
-Require Import Undecidability.Synthetic.MoreReducibilityFacts.
 
 Require Import Undecidability.TM.TM.
 
--- a/theories/MinskyMachines/MM2.v
+++ b/theories/MinskyMachines/MM2.v
@@ -11,7 +11,8 @@
 
 (** * Halting problem for two counter Minsky machines MM2_HALTING  *)
 
-Require Import List Arith Relations.
+Require Import Relations.Relation_Operators.
+#[local] Open Scope list_scope.
 
 Set Implicit Arguments.
 
--- a/theories/TRAKHTENBROT/Sig_rem_syms.v
+++ b/theories/TRAKHTENBROT/Sig_rem_syms.v
@@ -36,7 +36,7 @@ Section Sig_remove_symbols.
       and add an (interpreted) equality *)
 
   Definition Σnosyms : fo_signature.
-  Proof.
+  Proof using Σ.
     exists Empty_set (unit + (syms Σ + rels Σ))%type.
     + intros [].
     + intros [ | [ s | r ] ].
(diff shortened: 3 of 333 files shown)


mrhaandi:
For those who are using `vscoq`:
if `Proof using` is universal, then adding the following to `vscoq` settings makes for a nice up to 2x speedup in the editor:
```
"coqtop.args": ["-async-proofs-j", "2"],
```
This allows Coq to use two workers for asynchronous proof processing in the editor.
Useful for large files. Beware, more workers induce a larger memory footprint.

theories/_CoqProject
```diff
@@ -1,6 +1,7 @@
 -Q . Undecidability
 
 -arg -w -arg -notation-overridden
+-arg "-set 'Default Proof Using=Type'"
```
yforster, line 4:
```suggestion
-arg "-set \"Default Proof Using=Type\""
```
I need to change this line to process files in Proof General. Can you check whether it breaks something on your side again?

yforster (changes requested):
Apart from the one requested change, this is great!

theories/_CoqProject
mrhaandi: If I use `-arg "-set \"Default Proof Using=Type\""` as you suggest, then `make` does not compile the project anymore.

mrhaandi:
@DmxLarchey in your files I only added necessary `Proof using` annotations, so it should be fine right?

theories/H10/H10_undec.v
```diff
@@ -14,6 +14,7 @@ From Undecidability.Shared.Libs.DLW
 
 From Undecidability.Synthetic Require Import Undecidability.
 Require Import Undecidability.Synthetic.ReducibilityFacts.
+Require Import Undecidability.Synthetic.MoreReducibilityFacts.
```
DmxLarchey, line 17: @mrhaandi Do you remember why you need more `Import`here ?

theories/MinskyMachines/MM2.v
```diff
@@ -11,7 +11,8 @@
 
 (** * Halting problem for two counter Minsky machines MM2_HALTING  *)
 
-Require Import List Arith Relations.
+Require Import Relations.Relation_Operators.
```
DmxLarchey, line 14: @mrhaandi Ok this is a restriction. I hope the sub-module `Relations.Relation_Operators` name will remain stable over time.

theories/TRAKHTENBROT/Sig_rem_syms.v
```diff
@@ -402,14 +402,11 @@ Section completeness.
       rewrite fol_rels_are_functions_spec in HM; auto.
     Qed.
 
-    Let HA : fol_sem M φ (fol_rem_syms A).
```
DmxLarchey, line 405: @mrhaandi Since you removed this, may this was unused?

theories/TRAKHTENBROT/Sig_rem_syms.v
```diff
@@ -419,8 +416,11 @@ Section completeness.
         exact (fom_rels M (inr (inr r))).
     Defined.
 
+    Let HA : fol_sem M φ (fol_rem_syms A).
```
DmxLarchey, line 419: @mrhaandi Ok no ... just moved around sorry

theories/H10/H10_undec.v
mrhaandi: Possibly just `ReductionChainNotations`. I could move the notations module to `ReducibilityFacts` to avoid the extra import.

theories/MinskyMachines/MM2.v
mrhaandi: `Relations` (also `List` and `Arith`) contains definitions, lemmas, side-effects, etc.
Here only the definitions are necessary to define `MM2`. If something breaks (or some new side-effect messes things up) this helps localize the issue.

theories/H10/H10_undec.v
mrhaandi: I now moved `ReductionChainNotations` to `ReducibilityFacts` to avoid extra dependencies where possible.

DmxLarchey (approved):

Issue #150, opened 2022-06-02
Faster CI


mrhaandi:
First, looking at some CI runs in the PRs:
https://github.com/uds-psl/coq-library-undecidability/runs/6672738627
the `Build dependencies task` says:
```
 upgrade   coq                  8.15.0 to 8.15.dev
...
∗ installed coq.8.15.dev
...
 downgrade coq                  8.15.dev to 8.15.0
...
∗ installed coq.8.15.0
```
Is this dance a caching issue?

Second, I suggest to remove the `build-fast` task and add the `Build interfaces (vos)` task as a requisite for the `build-full` task.
This way, we have an early abort for the full compilation (not wasting CPU cycles).
The `Check interfaces (vok)` task did not turn out useful for us and is subsumed by `build-full`.

yforster:
Indeed!

PR #152, opened 2022-06-07
Trying to fix CI slowness problem

yforster:


mrhaandi:
What is your opinion on https://github.com/uds-psl/coq-library-undecidability/issues/150? In particular
> I suggest to remove the build-fast task and add the Build interfaces (vos) task as a requisite for the build-full task.
This way, we have an early abort for the full compilation (not wasting CPU cycles).
The Check interfaces (vok) task did not turn out useful for us and is subsumed by build-full.

Maybe this is the most efficient approach for the future Coq ci integration.
So if something is really broken, it gets noticed by `Build interfaces (vos)`, and otherwise only the necessary work is performed?

yforster:
Superseded by #157

PR #155, opened 2022-07-04
Equivalence proofs w.r.t. computability of relations

yforster:
This is a draft merging equivalence proofs for models of computation, i.e. where we talk about the computability of a functional relation rather than just halting problems.

It's needed to show that it doesn't matter which model of computation we use to define CT, and it's a strengthening of the DPRM theorem possibly needed for @dominik-kirst et al.'s formulation of an arithmetic form of CT.

I'm mostly filing this now to let @mrhaandi see what will remain needed, I'm worried that if #148 removes to many things it will become impossible to merge this code. I still plan to refactor files and put everything into a nice structure once the admits are filled.

There are some problems remaining:
- The definition of `SBTM` has changed, but is used in the proof. My plan would be to use the old definition of `SBTM` internally in the proof, but not expose somewhere that this alternative definition exists. @mrhaandi, do you have an opinion on this?
- There is an `admit` in `TM/L/CompilerBoolFuns/Compiler.v`. The lemma used there seems to not exist anymore? @fakusb  do you have an idea what's wrong? 
- There is an `Admitted` theorem in `MinskyMachines/MM/mm_comp_strong.v`. I had a slightly strengthened version of the compiler from BSM to MM, but with the refactoring of the compiler I don't understand what's going on anymore. @DmxLarchey could you have a look whether it's easy to prove the theorem `bsm_mm_compiler_strong`? 
- There is an `admit` in `L/Reductions/PCPb_to_HaltL.v`. I'll investigate myself, and if that doesn't work ask Fabian.


DmxLarchey:
Hi @yforster 

> There is an Admitted theorem in `MinskyMachines/MM/mm_comp_strong.v`. I had a slightly strengthened version of the compiler from BSM to MM, but with the refactoring of the compiler I don't understand what's going on anymore. @DmxLarchey could you have a look whether it's easy to prove the theorem bsm_mm_compiler_strong?

This file `MinskyMachines/MM/mm_comp_strong.v` is not part of the `coq-8.15` branch. Moreover, I do not remember writing that code. The statement and proofs in this file are not my style, eg `try firstorder congruence`. Is it possible that you wrote that code? In that case, could you please explain what was the issue with the original `BSM` to `MM` compiler that needed you to write that variant? 

Moreover, notice that there is now a `BSM(n)` to `MMA(1+n)` compiler in the code at `MinskyMachines/MMA/bsm_mma.v`.

yforster:
You didn't write that code, I did :) But now I can't merge it, because the compiler changed and I don't understand the new structure.

> In that case, could you please explain what was the issue with the original BSM to MM compiler that needed you to write that variant?

It's correctness statement was not strong enough for my use case. It seems like the problem was that the compiler inserted registers in the wrong order, the stronger version allows passing all registers as parameters (see the theorem). This was easily derivable from the generic compiler, the flexibility was just not used in the initial `mm_comp.v` proof.

yforster:
> Ok @yforster I will have a look at it or possibly rewrite that proof.

Thank you Dominique! :) 

With the help of Fabian I was able to fix the problem w.r.t. TMs and w.r.t. L, so now I just need to find a workaround for the SBTM problem.

DmxLarchey:
@yforster Done ;-) But may be you would like to clean the proof up a bit...

yforster:
Thank you Dominique! :)

After discussion with Andrej, we have the following proposal how to go on with this PR:

- I add the proof that `TM_computable -> L_computable` from the TM-invariance paper.
- Andrej adds a proof that `L_computable -> MMA_computable` (all infrastructure is there)
- Andrej adds a proof that `MMA_computable -> MM_computable` (should not be hard)
- Andrej adds a proof that `MMA_computable -> TM_computable` (should be more or less for free)

This avoids relying on the Hoare framework, with just minimal use of the complexity framework, and overall leads to a nice circle, with just TM as an outlier.

mrhaandi:
I also was too optimistic.
- `MMA_computable -> TM_computable` is not for free because currently `TM` cannot shrink its tape (e.g. from a tape content `1000` you cannot get `100` because you can only add or overwrite symbols (never remove).
- As for the `L_computable -> MMA_computable`, there are also more hurdles than anticipated. For one, `MMA` can simulate leftmost-outermost call-by-name really well, but `L` is call-by-value. For this I use Plotkin's CPS transformation, which again destroys natural input/output encodings.

Overall my suggestion for @yforster is as follows:
1) Clean up and refactor the current PR to have top-level `X_computable` definitions in the main files (e.g. `TM_computable` in `TM.v`, but not `TM_computable_iff` or other auxiliary lemmas, which go to `Util`).
2) As currently the case, put `X_computable_to_Y_computable.v` fies to corresponding `Reductions`.
3) Make sure that the code is robust (e.g. conformity with `Set Default Goal Selector "!"` and `Set Mangle Names`, and proper side effect localities).
4) Find a good place for the main computable implication loop theorem (maybe in top-level of `TM`). I suggest to put only the traditionally established models into the loop (`TM`, `MM(A)`, `H10`, `Mu`, `L`).

Then we can merge and I will see in the long run (possibly assisted by student thesis code) how alternative paths can result in faster/easier proofs for the main loop (this is the reason to have only the key models in the main loop). Of course, this will destructively interact with https://github.com/uds-psl/coq-library-undecidability/pull/148. I'll see what to do about it afterwards.

yforster:
If we put `X_computable` into `X.v`, then all `X.v` depend on Vectors. Are we fine with that?

yforster:
Unrelated: `Util` or `Utils`? :) We're not consistent in that regard. @DmxLarchey seems to prefer `Utils`, @mrhaandi `Util`, and I seem to randomly pick one of the two.

mrhaandi:
> Unrelated: `Util` or `Utils`?

Currently there are many `Util` folders but only two `Utils`. So former is the path of least resistance.

mrhaandi:
> If we put `X_computable` into `X.v`, then all `X.v` depend on Vectors. Are we fine with that?

If it is exactly `Vectors` (like `List`) from `Coq` and no further auxiliary files, which start to pollute a minimal definitions file, then it is fine. This is already the case in https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.15/theories/L/L.v

yforster:
I've reorganised the file structure and started a file `Synthetic/Models_Equivalent.v`. Everything works great, but something around `BSM_computable` is giving me a universe inconsistency :(

yforster:
Of course it's not something around `BSM_computable` in the end, but something in the Hoare framework. We've encountered that issue before, but were able to workaround. Now I don't see how to workaround... and I don't find a fix either. So for now, the big equivalence theorem has the `L -> TM` arrow commented out.

mrhaandi:
> We've encountered that issue before, but were able to workaround. Now I don't see how to workaround... and I don't find a fix either. So for now, the big equivalence theorem has the `L -> TM` arrow commented out.

More motivation to later use the `MMA` route.

yforster:
I inserted `Set Default Goal Selector "!".` into most (if not all) new files contributed by this PR. I won't be able to work on this considerably in the next months, so if there are no vetoes I would argue to merge soon. Maybe we manage to get an opam package before ITP/CoqWS with all new proofs?

mrhaandi:
> if there are no vetoes I would argue to merge soon. 

No vetoes. Hopefully we can fix the universes. I tracked it down to `vec_list_vec_map` in `PREP_spec_strong` possibly clashing with `WriteValue_SpecT_size`.

yforster:
I played around quite a lot with changing definitions in `WriteValue` and so on and was able to move the problem somewhere else, but in the end with no success. If you admit `PREP_spec_strong`, does that get rid of the problem?

yforster:
The error for sure is connected to the use of `eqType`. My approach to get rid of the issue was to disentangle that, and assume `eq_dec` in assumptions instead of `eqType`. That worked for a bit further, but then horrible things in tactics broke and I stopped following that route.

mrhaandi:
Possibly, I fixed the universes https://github.com/yforster/coq-library-undecidability/pull/2.
There was an issue with arguments for @DmxLarchey `vec_map` definition.

PR #157, opened 2022-07-04
Use `vos` compilation as fast-failing target before building `vo`, do not build `vok`

yforster:


mrhaandi:
To be ready for Coq 8.16 (see https://github.com/coq/coq/blob/v8.16/coq-core.opam which states `"ocaml" {>= "4.09.0"}`) please set OCaml to something newer (4.13 does work well).
Otherwise, feel free to merge anytime.

mrhaandi:
Also please squash commits before final merge to not have arbitrary wrong yml in the git history.

yforster:
I can't manage to use the `setup-ocaml` action with `flambda` enabled anymore, for new OCaml versions this only seems to be supported by the `setup-ocaml@v2` version. But that one automatically caches the OCaml build, in at least 2 problematic ways: There's no way to cache installed packages like MetaCoq, and there's no way to cache based on branches - at least I couldn't manage to. So back to `@v1`, with the expense of no `flambda`

PR #163, opened 2022-08-31
Added decider for reversible FRACTRAN halting

mrhaandi:
Based on code in the master's thesis by @archbung.
Small changes to the original code include:
- compatibility with `Set Default Goal Selector "!"`
- compatibility with `Set Mangle Names`
- explicit boolean decision procedure `decide : { P : list (nat * nat) | fractran_reversible P } * nat -> bool`, which is stronger than `exists f, ...`
- top-level `FRACTRAN_dec.v` result
- included @DmxLarchey as author because of remarks and ideas in an early private communication

Also fixed a copy-paste typo "autor", which is why many files are touched.

### DIFF
--- a/theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
+++ b/theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
@@ -1,0 +1,292 @@
+From Coq Require Import List Arith Lia Permutation Wf_nat.
+Import ListNotations.
+
+From Undecidability.Shared.Libs.DLW Require Import gcd prime.
+From Undecidability.FRACTRAN Require Import FRACTRAN fractran_utils.
+
+Set Default Goal Selector "!".
+
+(* necessary facts on prime factorization *)
+Module Prime_factors.
+
+#[local] Notation lprod := (fold_right mult 1).
+
+Lemma prime_divides_lt {lc ld} :
+  ~ (divides (lprod ld) (lprod lc)) ->
+  Forall prime lc -> Forall prime ld ->
+  exists (p : nat), prime p /\ count_occ Nat.eq_dec lc p < count_occ Nat.eq_dec ld p.
+Proof.
+  revert lc. induction ld as [|a ld IHld].
+  - intros lc H. exfalso. apply H.
+    exists (lprod lc). cbn. lia.
+  - intros lc H Hlc Hld'.
+    pose proof (Ha := Forall_inv Hld').
+    pose proof (Hld := Forall_inv_tail Hld').
+    destruct (in_dec Nat.eq_dec a lc) as [Halc|Halc].
+    + apply in_split in Halc as [l1c [l2c ?]]. subst lc.
+      assert (H' : not (divides (lprod (ld)) (lprod (l1c ++ l2c)))).
+      { intros [p Hp]. rewrite !lprod_app in *.
+        apply H. exists p. cbn. nia. }
+      assert (H'lc : Forall prime (l1c ++ l2c)).
+      { revert Hlc. rewrite !Forall_app.
+        now intros [? ? %Forall_inv_tail]. }
+      specialize (IHld _ H' H'lc Hld).
+      destruct IHld as [p [? Hp]].
+      exists p. split; [assumption|].
+      revert Hp. rewrite !count_occ_app. cbn.
+      destruct (Nat.eq_dec); lia.
+    + exists a. split; [assumption|].
+      apply (count_occ_not_In Nat.eq_dec) in Halc.
+      rewrite count_occ_cons_eq, Halc; lia.
+Qed.
+
+Definition count_pow (p n : nat) : nat :=
+  let (l, _) := (@prime_decomp (S n) (Nat.neq_succ_0 n))
+    in count_occ Nat.eq_dec l p.
+
+Lemma prime_divides_lt' c d (Hc : c <> 0) (Hd : d <> 0) :
+  ~ (divides d c) ->
+  exists p,
+    (forall x y (Hx : x <> 0) (Hy : y <> 0),
+      x * c = y * d -> count_pow p (pred y) < count_pow p (pred x)).
+Proof.
+  pose proof (@prime_decomp c Hc) as [lc [Hclc Hlc]].
+  pose proof (@prime_decomp d Hd) as [ld [Hdld Hld]].
+  subst. intros H. apply prime_divides_lt in H as [p [? Hp]]; [|assumption ..].
+  exists p. intros x y Hx Hy. unfold count_pow.
+  destruct (@prime_decomp (S (pred x)) _) as [lx [Hxlx Hlx]].
+  destruct (@prime_decomp (S (pred y)) _) as [ly [Hyly Hly]].
+  assert (S (pred x) = x) as H'x by lia.
+  assert (S (pred y) = y) as H'y by lia.
+  rewrite H'x in Hxlx. rewrite H'y in Hyly.
+  subst. rewrite <- !lprod_app.
+  intros E. apply prime_decomp_uniq in E; [|now rewrite Forall_app ..].
+  assert (iffLR : forall P Q, (P <-> Q) -> P -> Q) by tauto.
+  assert (H' := iffLR _ _ (Permutation_count_occ Nat.eq_dec _ _) E p).
+  rewrite !count_occ_app in H'. lia.
+Qed.
+
+Lemma rel_prime_intro {p q} : (p <> 0 \/ q <> 0) ->
+  { a & { b & { g | p = a*g /\ q = b*g /\ is_gcd a b 1 } } }.
+Proof.
+  intros ?. assert (Hg := gcd_spec p q).
+  assert (gcd p q <> 0).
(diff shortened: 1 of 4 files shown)


yforster:
In case you think that would be nice to have, there is https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors, which allows adding other people to a commit so GitHub keeps proper track of contributors.

DmxLarchey:
The new code is mostly orthogonal to the existing structure which is good. I am ok with the switch `Autor:` -> `Author(s):` but it has nothing to do with `FRACTRAN`. I don't ask to separate the PR in two halves but next time, I think it would be better to isolate such cosmetic changes in a separate PR which is easy and quick to review.

I will have a deeper look into the decision procedure hopefully soon.

mrhaandi:
> I am ok with the switch Autor: -> Author(s): but it has nothing to do with FRACTRAN. I don't ask to separate the PR in two halves but next time.

You are right, I'll just split the PR.

> which allows adding other people to a commit so GitHub keeps proper track of contributors.

This is a good idea, I did not know that. I'll change the commit message.

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
```diff
@@ -0,0 +1,292 @@
...
+
+(* necessary facts on prime factorization *)
+Module Prime_factors.
+
+#[local] Notation lprod := (fold_right mult 1).
+
+Lemma prime_divides_lt {lc ld} :
+  ~ (divides (lprod ld) (lprod lc)) ->
+  Forall prime lc -> Forall prime ld ->
+  exists (p : nat), prime p /\ count_occ Nat.eq_dec lc p < count_occ Nat.eq_dec ld p.
```
DmxLarchey, line 17: I guess this could be converted to `{ p : nat | ... }` instead of `exists`

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
```diff
@@ -0,0 +1,292 @@
...
+Qed.
+
+Definition count_pow (p n : nat) : nat :=
+  let (l, _) := (@prime_decomp (S n) (Nat.neq_succ_0 n))
+    in count_occ Nat.eq_dec l p.
+
+Lemma prime_divides_lt' c d (Hc : c <> 0) (Hd : d <> 0) :
+  ~ (divides d c) ->
+  exists p,
+    (forall x y (Hx : x <> 0) (Hy : y <> 0),
```
DmxLarchey, line 50: Why not formulate this as `(1+x)*c=(1+y)*d` and avoid `Hx`, `Hy` and `pred` altogether ?

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
```diff
@@ -0,0 +1,292 @@
...
+Import Prime_factors.
+
+Module Argument.
+
+Lemma fractran_nstop_zero_num_1 d l s : ~ (0,d) :: l /F/ s ↓.
+Proof.
+  destruct d as [|d]; simpl; intros [y [_ Hs]]; apply (Hs 0); now constructor.
+Qed.
+
+Lemma fractran_nstop_zero_num_2 a b d l x : ~ (S a, b) :: (0, d) :: l /F/ x ↓.
```
DmxLarchey, line 103: This could follow directly from if `l` never terminates then `(x,y) :: l` never terminates.

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
```diff
@@ -0,0 +1,292 @@
...
+  ((l2 /F/ n ↓) + (not (l2 /F/ n ↓))).
+Proof.
+  intros H [|Hn]; [left|right].
+  - eapply fractran_step_iff_halt; eassumption.
+  - intros ?. apply Hn. eapply fractran_step_iff_halt; [|eassumption].
+    firstorder easy.
+Qed.
+
+(* in a reversible FRACTRAN program the second fraction is shadowed by the first *)
+Lemma fractran_reversible_shadow {a b c d P x y} : x <> 0 -> y <> 0 ->
```
DmxLarchey, line 177: Perhaps use this to state (and prove) that if '(x,y) :: l' is a reversible fractran program then it is step-equivalent to '(x,y) :: nil'.

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
```diff
@@ -0,0 +1,292 @@
...
+    apply fractran_rt_no_zero_den in Hn.
+    - apply (Hy 0). apply in_fractran_0. lia.
+    - now repeat constructor. }
+  left. now apply fractran_stop_ndiv_singleton.
+Qed.
+
+(* informative decision statement for reversible FRACTRAN halting *)
+Theorem decision (P : list (nat * nat)) (n: nat) : fractran_reversible P -> (P /F/ n ↓) + (not (P /F/ n ↓)).
+Proof.
+  induction P as [P IH] using (induction_ltof1 _ (fun P => length P)); unfold ltof in IH.
```
DmxLarchey, line 234: I do think case analysis is enough on `P`, ie induction is not needed. Non-empty reversible fractran programs are step-equivalent to their single first fraction alone.

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
mrhaandi: `in_split` is used for the proof, which is propositional. So the `{ p : nat | ... }` is not immediate (and not necessary for the consequent proofs).

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
mrhaandi: Note that most of the code is written by a student. I merely made it so it does not immediately break on future Coq versions. I'd implement your simplification.

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
DmxLarchey: Btw, a fractran programs never terminates iff one of its fractions is an integer.

mrhaandi: I added `Lemma fractran_nstop_cons a b l s : (forall t, ~ l /F/ t ↓) -> ~ (a,b) :: l /F/ s ↓.` as a less ad-hoc lemma.

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
mrhaandi: Isn't induction necessary to prove the step-equivalence to the `(x,y) :: nil` form?
> Perhaps use this to state (and prove) that if (x,y) :: l is a reversible fractran program then it is step-equivalent to (x,y) :: nil.

This is the (inlined) argument of the decision procedure, for which induction is used.
Specifically, that in a reversible fractran program the first fraction shadows all other fractions.

DmxLarchey: > Isn't induction necessary to prove the step-equivalence to the (x,y) :: nil form?

Ok yes I think you are right. But indeed, that statement that _in a reversible fractran program the first fraction shadows all other fractions_ is not explicitly stated which blurs the argument for decision a bit.

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
mrhaandi: I now pulled out the following statement from the decision procedure for more clarity.
```
Lemma fractran_reversible_shorten {a b c d P s t} :
  fractran_reversible ((S a, b) :: (S c, d) :: P) ->
  ((S a, b) :: P /F/ s ≻ t) <-> ((S a, b) :: (S c, d) :: P /F/ s ≻ t).
```
If I were to consider arbitrary `(x,y) :: l` (instead of strictly positive numerators), then it results in a larger, unnecessary case analysis.

theories/FRACTRAN/Deciders/Halt_REV_FRACTRAN_dec.v
mrhaandi: It is now made more precise by `fractran_reversible_shorten` and the actual decision procedure is by induction/brief case analysis.

DmxLarchey:
Thx for considering the changes. The PR is ok for me now. I do not think I can formally approve your PR by an action in the GH interface. But I validate it. You can merge whenever you want.

mrhaandi:
> Thx for considering the changes. The PR is ok for me now. I do not think I can formally approve your PR by an action in the GH interface. But I validate it. You can merge whenever you want.

Thank you for your insights. The GitHub review interface is somewhat unintuitive I keep forgetting whether I have to push some button, or it is on your side in "files changed"/"review changes"/"approve".
I'll just merge then.

PR #165, opened 2022-09-01
replaced TM/Univ by top-level UTM construction

mrhaandi:
Universal, binary Turing machine with 1 tape
The construction is via a universal Krivine machine simulation.


mrhaandi:
Since there is an 8.16 branch, this PR is outdated. I'll mark it as draft for now.

PR #172, opened 2022-09-23
replaced parts of Shared/Libs/PSL by stdlib

mrhaandi:
Replaced `List` and `Vector` functions, which are already defined in Coq's standard library.


mrhaandi:
Compiles with latest published `metacoq 1.1+8.16`.
For the record, full compilation instructions are given below, since the CI is not ready yet.
```
opam switch create 8.16 ocaml-option-flambda
opam repo add coq-released https://coq.inria.fr/opam/released
opam update
opam pin add -n -k git coq-smpl.8.16 "https://github.com/uds-psl/smpl.git#coq-8.16"
opam install . --deps-only
make
```

PR #171, opened 2022-09-07
CI for Coq 8.16

yforster:

PR #173, opened 2022-10-12
replace prodLists with list_prod from stdlib

haansn08:


DmxLarchey:
Hi @haansn08. Is this a _systematic_ replacement of `prodList` by `list_prod` from StdLib?

DmxLarchey:
> @DmxLarchey Not sure what you mean by systematic, but I made sure the library still compiles.

I mean that there is no occurrence of `prodList` anymore after the resulting update? Or is this just a punctual change in two files ?

haansn08:
There are no other occurrences of `prodList`. I now also removed the Lemma `prod_nil` which also seems to be unused.

DmxLarchey:
Considering it is a systematic replacement with an equivalent tool from the standard library, I would likely ok the change. @yforster and @mrhaandi do you agree with merging this ?

DmxLarchey:
> There are no other occurrences of `prodList`. I now also removed the Lemma `prod_nil` which also seems to be unused.

DmxLarchey:
> > There are no other occurrences of `prodList`. I now also removed the Lemma `prod_nil` which also seems to be unused.

If `prod_nil` isn't already in the StdLib, maybe it could be left.

mrhaandi:
> > > There are no other occurrences of `prodList`. I now also removed the Lemma `prod_nil` which also seems to be unused.
> 
> If `prod_nil` isn't already in the StdLib, maybe it could be left.

`prod_nil` refers to the removed `prodList` and should be removed as well. Since the equivalent statement for `list_prod` is not necessary, there is no need to formulate it.

mrhaandi (approved):

mrhaandi:
I approve the proposed changes. @yforster it would be good to have a proper CI run before merge; what is the status of https://github.com/uds-psl/coq-library-undecidability/pull/171?

mrhaandi:
To be on the safe side (without currently functioning ci), I have compiled the PR locally without issues.

DmxLarchey (approved):

PR #174, opened 2022-10-13
Update README.md

haansn08:
To know this would have been helpful to me, as I run into [problems](https://github.com/MetaCoq/metacoq/issues/766) building the the "safechecker" part of the MetaCoq library. Thankfully coq-library-undecidability only requires the template-coq part which I could build successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ opam pin add coq-library-undecidability.dev+8.16 "https://github.com/uds-psl/coq
 
 ### Manual installation (TENTATIVE)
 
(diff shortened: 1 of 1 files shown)


mrhaandi:
The comment about Template-Coq should go to the Troubleshooting section of the readme.
It is not part of the recommended manual installation instructions.

mrhaandi:
> Hm, the comment clarifies which dependencies are needed to build the library rather than solving a specific problem a user might run into.

I see your point. The current formulation still may contribute to more confusion because of the overlapping conditional formulations for different manual installations approaches.
The preferred way is to install dependencies via opam. For this, the addition is not important and is potentially misleading.
Maybe, instead of the additional sentence, replace `MetaCoq` by "`Template-Coq` (part of `MetaCoq`)", which is most precise.

mrhaandi:
Btw. could you fix the MetaCoq link?
`https://metacoq.github.io/metacoq/` fails.
`https://metacoq.github.io` should work.

haansn08:
@mrhaandi I now fixed the link and squashed the commits.

PR #189, opened 2023-01-05
Draft: FOL Rework

JoJoDeveloping:
As part @dominik-kirst and my ongoing work on the FOL library, we wanted to re-organize the `FOL` subfolder here, so that you can base a library on it instead of having an unsorted mess of files.

This library mostly implements the approach already discussed, keeping a large part of the future FOL library separate from this library.

This MR completely reorganized the `FOL` subfolder. Unfortunately, the rework started off at a later stage, so some things need to be ported over again. Notably, the following commits affected the `FOL` subfolder between since the developments we intend to merge now diverged (the last common ancestor is `[5a6d81ed`](https://github.com/uds-psl/coq-library-undecidability/commit/5a6d81edb611039e8806a01e011cdd7f85f94fda) from April 12):

* [X] [cleaner ListAutomation](https://github.com/uds-psl/coq-library-undecidability/commit/63096ec79d8012dfd4202cb56cd12d41b37fa165)
  + all code compiles now after adding a few missing imports for `ListAutomationHints`, so I guess this is done
[X] [address Coq 8.16 warnings](https://github.com/uds-psl/coq-library-undecidability/commit/5bdd7b6b2ba8a4c907cd8cbea56a1b1e07da3dda)
  + all code that generated warnings was upgraded
[X] [Remove commented code](https://github.com/uds-psl/coq-library-undecidability/commit/aad76fd1acf7854354e1196edaa24eb55863d451)
  + the one affected code part was removed
[X] [improved goal and side effect management](https://github.com/uds-psl/coq-library-undecidability/commit/41c1311f45aaab04117d96fbf6841e6cf77aa929)
  + This commit also just changed imports originally
[X] [global Set Default Proof Using "Type"
    removed individual Default Proof Using
    refactored ReducibilityFacts for minimal dependencies](https://github.com/uds-psl/coq-library-undecidability/commit/2e66df2315d7969fb2fd5d9125d541c01c9c793e)
    + all individual `Set Default Proof Using "Type"` are removed
[ ] [established that intermediate files do not depend on _undec files
    do not unnecessarily Require Synthetic.Undecidability](https://github.com/uds-psl/coq-library-undecidability/commit/25f8d98a151233c473af4cd87c22aa31a63d8306)
     + this still needs to be addressed

Those refactorings need to be reviewed, and, if necessary, be re-applied. If done, this will be marked by a checked box.

Apart from this, some changes had to be made to files not in the FOL folder. These are:
* `PCP/PCP_undec.v` gained `Lemma PCP_compl_undec`, proving the undecidability of the complement of `PCP`, and similarly for `PCPb`
* `PCP/Util/PCP_facts.v` grew because some lemmas were moved into it, specifically that some sub-problems are decidable. These were previously part of the `TRAKHTENBROT` development, but are now also needed by new `FOL` results
* `SeparationLogic/Reductions/FSATdc_to_MSLSAT.v` was affected by changing imports for `FOL` files
* `Shared/ListAutomation.v` got new notations and tactics
* `Synthetic/DecidabilityFacts.v` got `decidable_iff'`, which is dual to `decidable_iff` (which is not actually an equivalence)
* `TRAKHTENBROT/bpcp.v` got leaner, as the aforementioned lemmas were moved from here.

Apart from this, more undecidability proofs are also contributed, namely those in `FOL/FSAT_direct_undec.v` (which re-use the reduction of [CPP'19](https://www.ps.uni-saarland.de/extras/fol-undec/)), and those in `FOL/FST_undec.v` and `FOL/binFST_undec.v`, due to @dominik-kirst and @HermesMarc as part of their [ITP'21 paper](https://www.ps.uni-saarland.de/extras/axiomatisations-ext/)


mrhaandi:
> `PCP/Util/PCP_facts.v` grew because some lemmas were moved into it, specifically that some sub-problems are decidable. These were previously part of the `TRAKHTENBROT` development, but are now also needed by new `FOL` results

If the results are of significance, then they could be separated into `Deciders` and `Enumerators` subfolders.
For example see [MinskyMachines/Deciders](https://github.com/uds-psl/coq-library-undecidability/tree/coq-8.16/theories/MinskyMachines/Deciders) or [L/Enumerators](https://github.com/uds-psl/coq-library-undecidability/tree/coq-8.16/theories/L/Enumerators).

> It might be prudent to squash all commits (which would lose authorship information).

Yes. It is possible to use the co-author feature for this.

> `Shared/ListAutomation.v` got new notations and tactics

(Only a remark) Except for `FOL` usage of this file is discouraged. Maybe at a later point it can move to `FOL` completely.

Also

- In the pull request the `FOLP` and `TRAKHTENBROT` folders are top level in `theories` (which is different from a prior discussion).
Will this be addressed in a later PR?

JoJoDeveloping:
> * In the pull request the `FOLP` and `TRAKHTENBROT` folders are top level in `theories` (which is different from a prior discussion).
>       Will this be addressed in a later PR?

This was an oversight. The `FOLP` folder has been made wholly redundant, the respective reduction/undecidability is now part of `FSAT_direct_undec.v`. It was supposed to be deleted, but merging with the 8.16 branch resurrected that folder.

The `TRAKHTENBROT` has now been moved to `FOL`.

JoJoDeveloping:
> > `PCP/Util/PCP_facts.v` grew because some lemmas were moved into it, specifically that some sub-problems are decidable. These were previously part of the `TRAKHTENBROT` development, but are now also needed by new `FOL` results
> 
> If the results are of significance, then they could be separated into `Deciders` and `Enumerators` subfolders. For example see [MinskyMachines/Deciders](https://github.com/uds-psl/coq-library-undecidability/tree/coq-8.16/theories/MinskyMachines/Deciders) or [L/Enumerators](https://github.com/uds-psl/coq-library-undecidability/tree/coq-8.16/theories/L/Enumerators).

I don't think they are of too much significance. Either is about 40 LoC. The file also contains useful inversion principles, as well as other useful small-scale lemmas.

JoJoDeveloping:
While checking that no `Require Synthetic.Undecidability` was forgotten, I noticed that some were forgotten in another folder. I have addressed this in #190.

DmxLarchey:
> The `TRAKHTENBROT` has now been moved to `FOL`.

Just a small remark (cc also @dominik-kirst).

The name `red_collection` loses the information that those reductions are targeted towards transfering undecidability.

Also, at the end, that file states the most general form of Trakhtenbrot we have, ie reduction from `BPCP` to `FSAT` for any signature which contain an at least binary symbol.

I would rather call it `red_chain_undec_complete` or `red_chain_undec_full`.

JoJoDeveloping:
> The name `red_collection` loses the information that those reductions are targeted towards transfering undecidability.

It's now called `red_chain_undec_full.v`

yforster (approved):

yforster:
@mrhaandi I added you as a reviewer just to ensure that you agree wity merging this in general with the current state to unblock other plans (such as integrating into Coq CI) - we probably might have minor suggestions that can be incorporated later.

I am unsure about the decision of squashing vs not squashing. Sure we can have a single squash commit with author information, but we lose information completely when which change was introduced and by whom. I also would be fine with an actual merge of the 200+ commits to keep this info.

mrhaandi (approved):

mrhaandi:
> I added you as a reviewer just to ensure that you agree wity merging this in general with the current state to unblock other plans (such as integrating into Coq CI) - we probably might have minor suggestions that can be incorporated later.

My earlier comments were considered and I approve of the current state.

I also am fine with the actual merge of the 200+ commits.

theories/Synthetic/Undecidability.v
```diff
@@ -22,8 +22,10 @@ Require Import Undecidability.TM.SBTM.
       PhD Thesis. Faculty of Mathematics and Computer Science of Saarland University. 2021.
       https://www.ps.uni-saarland.de/~forster/thesis.php
 *)
+Definition computational_explosion := enumerable (complement SBTM_HALT).
```
yforster, line 25: That's strange terminology. Where is it from?

JoJoDeveloping: Hm, that's a good question. I think @dominik-kirst and I came up with that name. It first appears in 36bcea9665ebf5d5834ff1b10279fd1b807ea163, where I seem to have introduced it.

JoJoDeveloping:
While working on the remaining library, I noticed that there were a few things FOL depends on that should have been migrated. One is the above ^^, the other is a useful lemma about reductions, and an enumerator for MuRec

theories/Synthetic/Undecidability.v
yforster: I'd argue for not including terminology that's not used in the underlying papers / theses. And I'd guess that the code is not affected by reverting this change, so it should be easy

yforster:
I'm happy with the enumerator and the Lemma. Regarding the latter at some point we might want to update / unify with the definitions from coq-synthetic-computability, but that's not urgent

theories/Synthetic/Undecidability.v
yforster: Besides that, I use "computational explosion" for `forall A : Type, False -> A` in my thesis, which seems to be natural in both directions for me -- it's the natural name for the principle, and if I hear the name, that's what I'd expect behind it because "explosion" is well-established for `forall A : Prop, False -> A`.

PR #199, opened 2023-05-05
Adapt w.r.t. coq/coq#17564.

ppedrot:
Should be backwards compatible.


DmxLarchey:
If I may, why do the `e` variants need to be introduced here (for the record).

yforster:
I think non-`e` tactics are not allowed to introduce evars by design. But there was a bug which allowed us to use `intros` in way that it was generating evars via `%` - now such uses have to be explicit via `e`. Did I get this right @ppedrot?

ppedrot:
Indeed, in Coq master intropatterns can silently introduce evars even though the surrounding tactic should prevent them. This happens when destructing higher-order terms.

ppedrot:
(You can merge this already since the CI is green.)

PR #201, opened 2023-06-12
Fix CI

yforster:


yforster:
I can't get caching to work atm, but 25 minutes of CI time seem okay, so I'm merging this now so we at least have working CI

PR #205, opened 2023-08-23
Add Intersection Type related problems

mrhaandi:
New problems
- strong normalization wrt. beta-reduction in the lambda-calculus
- intersection type inhabitation, typability, type checking

New reductions
- reduction from Krivine machine halting to strong normalization
- reduction from intersection type typability to intersection type type checking
- reduction from strong normalization to intersection type typability
- reduction from simple semi-Thue system 01 rewriting to intersection type inhabitation

### DIFF
--- a/theories/LambdaCalculus/Lambda.v
+++ b/theories/LambdaCalculus/Lambda.v
@@ -7,8 +7,8 @@
 
 (*
   Problem(s):
-    Weak call-by-name leftmost outermost normalization for closed lambda-terms (wCBNclosed)
-    Strong normalization for closed lambda-terms (SNclosed)
+    Weak call-by-name leftmost outermost normalization for given closed lambda-terms (wCBNclosed)
+    Strong normalization for given closed lambda-terms (SNclosed)
 
   Literature:
     [1] Plotkin, Gordon.
@@ -49,14 +49,14 @@ Fixpoint subst (sigma: nat -> term) (s: term) : term :=
 
 Notation closed t := (forall (sigma: nat -> term), subst sigma t = t).
 
-(* beta-reduction *)
+(* beta-reduction (strong call-by-name reduction) *)
 Inductive step : term -> term -> Prop :=
-  | stepSubst s t   : step (app (lam s) t) (subst (scons t var) s)
+  | stepSubst s t   : step (app (lam s) t) (subst (scons t (fun x => var x)) s)
   | stepAppL s s' t : step s s' -> step (app s t) (app s' t)
   | stepAppR s t t' : step t t' -> step (app s t) (app s t')
   | stepLam s s'    : step s s' -> step (lam s) (lam s').
 
-(* strong normalization of closed lambda-terms *)
+(* given a closed lambda-term s, is it strongly normalizing wrt. beta-reduction? *)
 Definition SNclosed : { s : term | closed s } -> Prop :=
   fun '(exist _ s _) => Acc (fun x y => step y x) s.
 
@@ -65,6 +65,6 @@ Inductive wCBN_step : term -> term -> Prop :=
   | wCBN_stepSubst s t  : wCBN_step (app (lam s) t) (subst (scons t var) s)
   | wCBN_stepApp s s' t : wCBN_step s s' -> wCBN_step (app s t) (app s' t).
 
-(* weak call-by-name leftmost outermost normalization of closed lambda-terms *)
+(* given a closed lambda-term s, is it normalizing wrt. call-by-name leftmost outermost reduction? *)
 Definition wCBNclosed : { s : term | closed s } -> Prop :=
   fun '(exist _ s _) => exists t, clos_refl_trans term wCBN_step s (lam t).
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ An equivalence proof that most of the mentioned models of computation compute th
 - Uniform boundedness of deterministic, length-preserving stack machines (`SMNdl_UB` in [`StackMachines/SMN.v`](theories/StackMachines/SMN.v))
 - Semi-unification (`SemiU` in [`SemiUnification/SemiU.v`](theories/SemiUnification/SemiU.v))
 - Halting problem for Krivine machines (`KrivineM_HALT` in [`LambdaCalculus/Krivine.v`](theories/LambdaCalculus/Krivine.v))
-- Strong normalization wrt. beta-reduction in the lambda-calculus (`SNclosed` in [`LambdaCalculus/Lambda.v`](theories/LambdaCalculus/Lambda.v))
+- Strong normalization for given terms in the strong call-by-name lambda-calculus (`SNclosed` in [`LambdaCalculus/Lambda.v`](theories/LambdaCalculus/Lambda.v))
 - System F Inhabitation (`SysF_INH` in [`SystemF/SysF.v`](theories/SystemF/SysF.v)), System F Typability (`SysF_TYP` in [`SystemF/SysF.v`](theories/SystemF/SysF.v)), System F Type Checking (`SysF_TC` in [`SystemF/SysF.v`](theories/SystemF/SysF.v))
 - Intersection Type Inhabitation (`CD_INH` in [`IntersectionTypes/CD.v`](theories/IntersectionTypes/CD.v)), Intersection Type Typability (`CD_TYP` in [`IntersectionTypes/CD.v`](theories/IntersectionTypes/CD.v)), Intersection Type Type Checking (`CD_TC` in [`IntersectionTypes/CD.v`](theories/IntersectionTypes/CD.v))
 
--- a/theories/IntersectionTypes/Reductions/SNclosed_to_CD_TYP.v
+++ b/theories/IntersectionTypes/Reductions/SNclosed_to_CD_TYP.v
@@ -7,7 +7,7 @@
 
 (*
   Reduction from:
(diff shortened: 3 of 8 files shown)


theories/LambdaCalculus/Lambda.v
```diff
@@ -0,0 +1,70 @@
...
+Notation closed t := (forall (sigma: nat -> term), subst sigma t = t).
+
+(* beta-reduction *)
+Inductive step : term -> term -> Prop :=
+  | stepSubst s t   : step (app (lam s) t) (subst (scons t var) s)
+  | stepAppL s s' t : step s s' -> step (app s t) (app s' t)
+  | stepAppR s t t' : step t t' -> step (app s t) (app s t')
+  | stepLam s s'    : step s s' -> step (lam s) (lam s').
+
+(* strong normalization of closed lambda-terms *)
```
DmxLarchey, line 59: Wouldn't _strongly normalizable_ be more accurate? Strong normalization is a property of a calculus, not of individual terms?

theories/LambdaCalculus/Lambda.v
```diff
@@ -0,0 +1,70 @@
...
+  match s with
+  | var n => sigma n
+  | app s t => app (subst sigma s) (subst sigma t)
+  | lam s => lam (subst (scons (var 0) (funcomp (ren S) sigma)) s)
+  end.
+
+Notation closed t := (forall (sigma: nat -> term), subst sigma t = t).
+
+(* beta-reduction *)
+Inductive step : term -> term -> Prop :=
```
DmxLarchey, line 53: Is there an intent to refrain from using (infix) notations for beta-steps, substitutions etc? Even just locally?

DmxLarchey:
Btw about ten years ago, I did implement a [significant part of Krivine's book](https://members.loria.fr/DLarchey/files/Lambda_Calculus/index.html) in Coq, including in particular equivalence between SN and typability in D of pure lambda terms. The project used scoped De Bruijn syntax though.

I did use the following additional reference for system D: _Strict Intersection Types for the Lambda Calculus_ by Steffen van Bakel. Empty intersections were also forbidden but by assuming that the list was not empty instead of representing it as a pair of the head and a tail. 

For some reason I do not recall now, I had to specialize the type system to bounded assignments, ie measure the height of typing derivations. I guess some proofs relied on that measure. Do you also need it for your purpose of just characterizing SN with D-typability?

mrhaandi:
> Do you also need it for your purpose of just characterizing SN with D-typability?

No, there are several key aspects to this particular proof.
- As pointed out in [1, Remark 17.2.16 (ii)] almost all published proofs of this fact (including van Bakel) contain mistakes. Chances are, you had exactly this problem. I also had this issue.
- The technique I use is not exactly from literature. Instead of using strong normalization directly, I rely on a family of perpetual reduction strategies. It is very similar to [2], however I do not use a fixed reduction strategy ([2] uses Barendregt's perpetual F_infinity reduction strategy). This results in a simpler proof. 

[1] Barendregt, Henk, Wil Dekkers, and Richard Statman. Lambda calculus with types. Cambridge University Press, 2013.
[2] Neergaard, Peter Møller. "Theoretical pearls: A bargain for intersection types: a simple strong normalization proof." Journal of Functional Programming 15.5 (2005): 669-677.

theories/LambdaCalculus/Lambda.v
mrhaandi: I rephrased the decision problem as a question:
"given a closed lambda-term s, is it strongly normalizing wrt. beta-reduction?"
In the readme I now say (since I want to use a noun as the problem name) 
"Strong normalization wrt. beta-reduction for given terms in the lambda-calculus"
I would not like to say e.g.
"Strongly normalizable wrt. beta-reduction in the lambda-calculus"
because this is an adjective.

theories/LambdaCalculus/Lambda.v
mrhaandi: I presume you mean something like
```coq
#[local] Reserved Notation "s →β t" (at level 70).

(* beta-reduction *)
Inductive step : term -> term -> Prop :=
  | stepSubst s t   : app (lam s) t →β subst (scons t var) s
  | stepAppL s s' t : s →β s' -> app s t →β app s' t
  | stepAppR s t t' : t →β t' -> app s t →β app s t'
  | stepLam s s'    : s →β s' -> lam s →β lam s'
where "s →β t" := (step s t).

(* given a closed lambda-term s, is it strongly normalizing wrt. beta-reduction? *)
Definition SNclosed : { s : term | closed s } -> Prop :=
  fun '(exist _ s _) => Acc (fun x y => y →β x) s.
```

In fact, I find this difficult to parse (especially due to infix notations) compared to the notation-free version.
I also prefer (similar to `L.v`) to keep the definition mostly notation-free (maybe plus some standard library).
Interacting with the library, I sometimes get confused with all `P // e ▷ v`, `ρ // x ⇢ y`, `P // r ↠ s`, `ρ // s -1> t`, ...

DmxLarchey: Yes indeed something like this. Plus notations for lambda terms: application, abstraction, substitution of the variable 0. 

In particular, even thought I am used to it, I still felt that `subst (scons t var) s` was hard to parse for `s[t]`. Possibly something like 

```
£ x := var x
t ::ₛ σ := scons t σ
s{σ} := subst σ s
s[t] := s{t ::ₛ fun x => £x}
```

Using the constructor `var` for `fun x => var x` is confusing I think. The definition of the type of lambda terms is also imported so the name `var` itself does not necessarily lead to an obvious immediate interpretation.

I agree that if the definitions of notations are not visible close to where they are used, it can lead to difficult parsing.

However, a notation similar to usual mathematical prose can also be more convincing than standard prefix notation which non-CS users might feel hard to parse. Of course this is a matter of taste...

DmxLarchey:
@mrhaandi Concerning the equivalence between `SN` and typability in system D also denoted `λ∩⁻` in the online version of [Lambda Calculus with Types](https://home.ttic.edu/~dreyer/course/papers/barendregt.pdf) [1]. Of course I first had to realize that `SN` had to be defined inductively in terms of the `Acc` predicate, and especially, not in terms of infinite reduction sequences (as done in most textbooks that do not focus on constructive reasoning).

For both sides of the equivalence, I did (try) to follow the proof in [Krivine's book](https://www.irif.fr/~krivine/articles/Lambda.pdf) [3]:

- to show that every D-typable term is `SN`, the proof follows the technique of Girard, also reproduced in [1] as I understand. That proof did work out except that the measure I use to prove `SN a -> SN (u[a] t1 ... tn) -> SN ((λu) a t1 ... tn)` is not `N(t) := the sum of the length of all the normalizations of t` (as suggested in the book), but `L(t) := the length of the longest normalisation of t` instead. 
- For the converse, ie every `SN` term is D-typable, the proof in Krivine's book fails (both in the several textbook versions and in the online version [3]). Indeed the induction measure used to prove Thm 4.17 pp 69-70, namely `N(t)` above again, is actually not decreasing in the case of `t := λ...λ ((x) t1)` where `x` is a variable. Indeed, in that case, there is a length-preserving one-to-one correspondence between the normalization of `t` and the normalizations of `t1` hence we have `N(t) = N(t1)` contrary to what is stated in the book. Notice that, and this is misleading, the argument works for `t := λ...λ ((x) t1 t2)` or above. To solve the issue, I did replace the measure by a lexicographic product of `L(t)` (as above) and strict subterm, and it did work out.

There is some work to define `L(t)` by `Fixpoint` on `SN t`, so dependent on `sn : SN t` but _irrelevant in that proof_. This relies on the fact that beta-reduction steps are finitely branching. I would consider it as easy exercise now, but for me at that time, it was not so easy to work with the dependent version of `Acc_rect`.

Btw, I cannot find remark 17.2.16 in that online version of [1] and I do not have access to the paper book. Could you please forward me the text of the remark please? I am not sure the issue and your solution is the same as the one I did encounter.

theories/LambdaCalculus/Lambda.v
mrhaandi:
```coq
£ x := var x
t ::ₛ σ := scons t σ
s{σ} := subst σ s
s[t] := s{t ::ₛ fun x => £x}
```
autosubst uses something similar. I had a very confusing experience with it (and it did clash with both the standard library and with ssreflect).
> Of course this is a matter of taste...

I strongly prefer to keep it as is.

> Using the constructor `var` for `fun x => var x` is confusing I think.

I agree. I changed `(scons t (fun x => var x))` to `(scons t (fun x => var x))`.

mrhaandi:
> For the converse, ie every `SN` term is D-typable, the proof in Krivine's book fails (both in the several textbook versions and in the online version [3]).

I studied several approaches. There are three different textbook ones, Krivine's is broken, but can be fixed (as you say). Maybe the simplest proof is in the printed version of Lambda Calculus with Types (2013).
I went for my own (which is different from the existing ones) due to ongoing work on a different calculus for which only this approach works. I plan to refer to the implementation in PR as an additional use case of the technique.

yforster (approved):

mrhaandi:
> This can indeed be considered the lambda-calculus, but we use the terminology weak call-by-name, so we might as well be precise here.

@yforster I changed the documentation and comments from
`Strong normalization wrt. beta-reduction for given terms in the lambda-calculus`
to
`Strong normalization for given terms in the strong call-by-name lambda-calculus`

DmxLarchey (approved):

Issue #207, opened 2023-09-29
Consider changing to a permissive or weak copyleft license


palmskog:
To enable easier reuse of this library, please consider changing its license to a permissive license like [MIT](https://spdx.org/licenses/MIT.html) or a weak copyleft license like [MPL-2.0](https://spdx.org/licenses/MPL-2.0.html).

The issue with strong copyleft such as CeCILL-2.1 is that clients of the library will need to change their licenses to a strong copyleft license like CeCILL-2.1 itself or GPL, which they may not be prepared to do.

Changing this library to a permissive or weak copyleft license is (at least currently) a prerequisite for solving addressing https://github.com/uds-psl/coq-library-fol/issues/2

yforster:
I'm in favour of changing the library to the weakest license possible: MIT. In the end, we want as many people as possible to use our library :) @DmxLarchey @mrhaandi, do you have opinions on this?

palmskog:
@DmxLarchey CeCILL-B is definitely a step in the right direction for reuse compared to CeCILL-2.1, but the problem with it is that it's **incompatible** with GPL, as stated by the Free Software Foundation [themselves](https://www.gnu.org/licenses/license-list.html#CeCILL-B):

> The CeCILL-B is a free software license. It is incompatible with the GPL because it has requirements that are not present in the GPL. The credit requirements in section 5.3.4 exceed those of the GPL. It also has a strange requirement that you use your “best efforts” to get third parties to agree “to comply with the obligations set forth in this Article.” Please do not release software under this license.

What this means in practice is that if someone has library under say GPL-2 or GPL-3 and they use your library, then they can't legally distribute the resulting `.vo` files. You can see some advice from @zimmi48 on CeCILL-B licensing for another project [here](https://github.com/coq-community/apery/issues/10).

Zimmi48:
>CeCILL-B is definitely a step in the right direction for reuse compared to CeCILL-2.1

Given the reason that you provide in the rest of your message, I think we could even debate if CeCILL-B would really represent a step in the right direction.

[...]

PR #208, opened 2023-10-03
Change license from CeCILL v2 to MPL-2.0 (more permissive)

DmxLarchey:
Following the issue #207, I propose this relicensing PR.

No functional change at all here since only _at most one line of comment_ is updated for each `.v` file. 

However, please check that the two added files `MPL-2.0_LICENSE.txt` and `scripts/relicense_to_MPL.txt` are properly placed. 

Possibly we could consider a `licenses` subdir? Also about the new `scripts` subdir:
- is it properly placed? named?
- it is here only to remind what was done.

The `CeCILL_LICENSE.txt` file is kept for the moment but could be pruned easily in a later update.


Zimmi48:
Usually, when a project is not dual-licensed, it is recommended to keep a single license file and to name it `LICENSE`. This helps both humans and automated tools understand more quickly what the project license is.

For instance, GitHub will automatically display the license it detected in this summary of the repository:

image

Note that you should also update the `license` field in the `opam` file.

yforster:
Maybe we leave the commit as is, and put a commit on top that removes the CECILL file and the `scripts` subdir? That way the information is not lost because it strays part of the git tree, but it also stays around too prominently

DmxLarchey:
@Zimmi48 thx for the `opam` tip ...

DmxLarchey:
Ok so I squash merge this PR and then submit another one that removes the reference to the former CeCILL license and the `scripts` subdir.

mrhaandi (approved):

PR #220, opened 2024-04-04
Higher-order beta-matching

mrhaandi:
- added higher-order matching wrt. beta-equivalence `HOMbeta` 
- added reduction from `SSTS01` to `HOMbeta` resulting in `undecidable HOMbeta`

The reduction is quite different from the known technique by Loader [1].
It is based on the technique used for intersection type inhabitation, see `IntersectionTypes/Reductions/SSTS01_to_CD_INH.v`, sharing a common seed. A write up for a potential publication is in progress.

[1] Loader, Ralph. "Higher order β-matching is undecidable." Logic Journal of the IGPL 11.1 (2003): 51-68.


yforster (approved):
I think this looks great, especially the `HOMatching.v` files is exemplary for clearness of a problem introduction

PR #226, opened 2024-10-01
reductions to `HaltLclosed` and `L_computable_closed` without the `L` framework

mrhaandi:
- added reduction from `MMA_HALTING` to `HaltLclosed` without the `L` framework
- added `MMA_computable` to `L_computable_closed` without the `L` framework
- added `MM_computable` to `MMA_computable` reduction to close the loop
- made `HaltLclosed` more prominent in reduction chains
- generalized deterministic simulation to uniformly confluent simulation (for `L` reductions)

Now we have no `_undec` results which depend on the `L` framework / MetaCoq.
The only computability result which depends on the `L` framework / MetaCoq is `MuRec_computable_to_L_computable`.


yforster (approved):
I have no strong opinion, but I'm also not opposed, this indeed probably makes long-term maintainability easier. I'd not like to drop the reduction relying on the framework completely from the code to keep the framework alive (but if I'm not missing something this PR is not deleting the reduction anyway, it's just making sure the main loop does not rely on the framework)

mrhaandi:
> (but if I'm not missing something this PR is not deleting the reduction anyway, it's just making sure the main loop does not rely on the framework)

Yes, this PR just adds different options and does not delete reductions.

Issue #228, opened 2025-01-17
Notation Warnings


mrhaandi:
@DmxLarchey In the `master` branch, I suppressed some notation warnings in your code via `-closed-notation-not-level-0,-postfix-notation-not-level-1`.
Could you verify that suppressing the warning will not break the code after the deprecation deprecation phase? Or (if possible) fix the underlying issues?

DmxLarchey:
Hi @mrhaandi, is `master` supposed to be compiled with `rocq` only ?

DmxLarchey:
See #229

mrhaandi:
> is `master` supposed to be compiled with `rocq` only ?

Currently it requires `rocq` as the minimum. It would be nice to keep it that way for maximal compatibility and long-term maintenance.
Of course, you if you have the `coq` layer installed in addition (for development / debugging) it should also compile.

mrhaandi:
@yforster I did not touch the CI (which currently fails and needs a corresponding adjustment for `rocq` compilation).

mrhaandi:
I have created a `9.0` branch and fixed the L framework in that branch, so it compiles (without warnings) with MetaRocq.

PR #229, opened 2025-01-20
Test PR related to Issue 228

DmxLarchey:
See #228


theories/Shared/Libs/DLW/Vec/vec.v
```diff
@@ -539,7 +539,7 @@ Arguments vec_zero {n}.
 Arguments vec_one {n}.
 
 Reserved Notation " e '#>' x " (at level 58, format "e #> x").
-Reserved Notation " e [ v / x ] " (at level 57, v at level 0, x at level 0, 
+Reserved Notation " e [ v / x ] " (at level 1, v at level 0, x at level 0,
```
DmxLarchey, line 542: The change of priority here (or else in file `theories/MinskyMachines/MMenv/env.v`) results in messing with the SSR syntax for `rewrite`, see eg file https://github.com/uds-psl/coq-library-undecidability/blob/33c223126a65ec8db577412e7aab5737209df865/theories/MinskyMachines/Deciders/MM_2_HALTING_dec.v#L53

mrhaandi: Is it due to [unscoped, global, top-level notations](https://github.com/coq/coq/wiki/Anti-patterns#Unscoped-global-top-level-notations)? If so, is it much work to change that to opt-in notations (a common design practice for modular code)?

theories/Shared/Libs/DLW/Vec/vec.v
```diff
@@ -538,12 +538,18 @@ Arguments vec_plus {n}.
 Arguments vec_zero {n}.
 Arguments vec_one {n}.
 
-Reserved Notation " e '#>' x " (at level 58, format "e #> x").
-Reserved Notation " e [ v / x ] " (at level 57, v at level 0, x at level 0, 
+Module vec_notations.
```
DmxLarchey, line 541: Created a `Module` to be imported for these notations instead of redefining them over and over again. The notations are not globally `Reserved` anymore.

DmxLarchey: This creates issues mostly with:
1. the SSR syntax of the tactic `move => [ ]` especially when either when nesting `[]` or with successive `[]`. My guess is that `[` and `]` are not paired properly anymore;
2. the `nil`/`[]` or singleton `_::nil`/`[_]` notations that overlap between `ListNotations` and `vec_notations`. 

I did correct the issues in proofs scripts by splitting `move=>` into several successive moves and replacing `[]` with `nil`.

theories/Shared/Libs/DLW/Vec/vec.v
DmxLarchey: I modified the code to have a module `vec_notations` but it created some issues, see below. @mrhaandi does it suit you, and are the corrections it implied satisfactory for you ?

DmxLarchey:
The last warning issue remaining is related to 
https://github.com/DmxLarchey/coq-library-undecidability/blob/7b03998b700bd304d8e0480881d0cb6f8f57e078/theories/MinskyMachines/MM2.v#L45

```coq
Reserved Notation "i '//' r '⇢' s" (at level 70, no associativity).
Reserved Notation "P '//' r '→' s" (at level 70, no associativity).
Reserved Notation "P '//' r '↠' s" (at level 70, no associativity).
Reserved Notation "P '//' r ↓" (at level 70, no associativity).
```

where Coq/Rocq asks for the reserved notation  `_ // _  ↓` to be put at level 1 (instead of 70), because it ends with a terminal symbol. However, doing so has the side effect of preempting other notations like `_ // _  ⇢ _` which do not end with a terminal symbol. Hence, if they are not also at level 1, this mainly forbids the parsing of the other notations.

The reason I did choose level 70 for these is logical: these are `Prop` atoms and this level is nice for those kinds of terms.

@mrhaandi I do not know how to handle this at the moment?.

DmxLarchey:
The last commit 10478abd28384104bd6a6b024e8798ba275a485a solves the issue of left/right bi-closed notations that should be at level 0 according to Coq/Rocq. These had no side effects at all, so a straightforward correction.

mrhaandi:
@DmxLarchey  Is it correct that the last remaining issue is:
```
File "./Shared/Libs/DLW/Code/sss.v", line 25, characters 0-65:
Warning: Postfix notations (i.e. starting with a nonterminal symbol and
ending with a terminal symbol) should usually be at level 1 (default).
[postfix-notation-not-level-1,parsing,default]
```

DmxLarchey:
@mrhaandi Yes this is correct except that the very same issue also pops up in:

https://github.com/uds-psl/coq-library-undecidability/blob/70dfc56f33a6e4835281044e68aa68279989047e/theories/MinskyMachines/MM2.v#L45-L48

One possibility could be to modify the notation for "P terminates on input t" to something like `P ↓ t` but it would be quite dissimilar to the other forms `P // ...`.

mrhaandi:
If the final one is too ugly / bothersome to fix, we can for the near future use the same local warning suppression as the Coq standard library:
https://github.com/coq/coq/blob/c1d8bee3b89b2fc9ba6c76f472983e0f2d194f75/theories/ssr/ssreflect.v#L98

DmxLarchey:
Thx for the advice @mrhaandi !!

mrhaandi (approved):

mrhaandi:
> If you are ok, I can merge this PR corresponding to issue #228

Feel free to merge.
If you like, you can also inspect the `-notation-incompatible-prefix` warnings in your code (which currently are also suppressed) as part of another PR.

DmxLarchey:
>  you can also inspect the `-notation-incompatible-prefix` warnings in your code (which currently are also suppressed) as part of another PR.

@mrhaandi task completed in PR #230

PR #233, opened 2025-02-05
move notations into module

JoJoDeveloping:
This notation is reserved and only used in some parts of the development. But since it is still globally reserved, importing `vec.v` still produces warnings about clashing notations. A proper fix is to move these notations into a module. 

This is a first draft, presumably the definition should also be moved into such a module so that one does not have to re-define the notation elsewhere.

Making the notation no longer globally enabled would be great because it otherwise clashes with the ones defined in [FOL/Syntax/SyntacticOps.v](https://github.com/uds-psl/coq-library-undecidability/blob/coq-8.20/theories/FOL/Syntax/SyntacticOps.v).

### DIFF
--- a/theories/FOL/TRAKHTENBROT/discrete.v
+++ b/theories/FOL/TRAKHTENBROT/discrete.v
@@ -14,6 +14,7 @@ From Undecidability.Shared.Libs.DLW.Utils
 
 From Undecidability.Shared.Libs.DLW.Vec 
   Require Import pos vec.
+Import VectorNotations.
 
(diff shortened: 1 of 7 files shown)


DmxLarchey:
@JoJoDeveloping Please have a look at PR #229 that does precisely that among other things. It was only merged in `master` though.

JoJoDeveloping:
Oh, thanks for the hint. That seems to do the trick, sad that it's not available on 8.20 yet. So I guess we just wait until you're updated to 9.0.

PR #232, opened 2025-01-24
Cleanup for overridden notations

DmxLarchey:
This PR is a follow up of #231. It intends to solve remaining issues related to overridden notations, ie warnings that pop up when compiling without the flag `-notation-overridden`. @dominik-kirst would you mind having a look? Orelse pinging the author of the corresponding files? I also ping @mrhaandi who initiated those warning issues in the `master` branch.

Notice that this branch is supposed to be compiled with `rocq`, not `coq`.

The concerned files are below.

Proposed solutions by @DmxLarchey :
- `FOL/FOL.v`;
- `FOL/Utils/FriedmanTranslation.v`;
- `FOL/Utils/FriedmanTranslationFragment.v`;
- `FOL/Reductions/H10p_to_FA.v`.

Proposed solutions by @mrhaandi :
- `FOL/binFST_undec.v`;
- `FOL/Semantics/FiniteTarski/DoubleNegation.v`;
- `FOL/Reductions/H10UPC_to_FOL_friedman.v`;
- `FOL/Reductions/binZF_to_binFST.v`;
- `FOL/Reductions/ZF_to_FST.v`.

PR #236, opened 2025-02-24
Remove ill-placed notation definitions from FOL Friedman translation

JoJoDeveloping:
Turns out that putting these notations here is a bad idea for coq-library-fol. In particular, `-fol` likes to tweak these notations in various places which leads to overridden notation warnings. Fixing them is mostly downstream but requires that these notations are removed here.

### DIFF
--- a/theories/FOL/Utils/FriedmanTranslation.v
+++ b/theories/FOL/Utils/FriedmanTranslation.v
@@ -8,10 +8,6 @@ From Undecidability.FOL Require Import FullSyntax.
 From Undecidability.FOL.Arithmetics Require Import Robinson NatModel.
 Import Vector.VectorNotations.
 
-Notation "I ⊨= phi" := (forall rho, sat I rho phi) (at level 20).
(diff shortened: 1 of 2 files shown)


JoJoDeveloping:
the CI failure is not because of me, but because the docker container's user is now called `rocq` and not `coq`.

mrhaandi:
Is it possible to coordinate this with https://github.com/uds-psl/coq-library-undecidability/pull/232, e.g. as a pull request for [DmxLarchey:notation-overridden-2](https://github.com/DmxLarchey/coq-library-undecidability/tree/notation-overridden-2)?
@JoJoDeveloping feel free to review / comment on https://github.com/uds-psl/coq-library-undecidability/pull/232.

JoJoDeveloping:
Pr #232 also fixes this issue, but it might still be good to remove these notations here anyway since they are not needed.

mrhaandi:
Integrated in https://github.com/uds-psl/coq-library-undecidability/pull/232.

PR #249, opened 2026-06-12
Mention problems on CFG in README

meithecatte:
I noticed that the README didn't mention the CFG problems, even though I saw them in the library – so I added them.

Open to feedback on the phrasing and structure.


DmxLarchey:
> Also, I'm having trouble finding any literature on the "Post grammars", is this standard terminology?

@yforster could we have your input on this?

PR #250, opened 2026-06-13
Bump the recommended OCaml version to 4.14.2

meithecatte:
The 4.14.2 patch release incorporates ocaml/ocaml#12577, which makes the toolchain compile without errors on new versions of GCC. This is already necessary to be able to set up the library on bleeding-edge systems, such as Arch Linux, and the requirement will spread as the GCC update trickles down across distributions.

The CI is already set up to use the latest available OCaml release from the 4.14 branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ You can use `opam` to install the current state of this branch as follows.
 We recommend creating a fresh opam switch:
 
 ```
-opam switch create coq-library-undecidability --packages=ocaml-variants.4.14.1+options,ocaml-option-flambda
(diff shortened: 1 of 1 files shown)


DmxLarchey:
Dear @meithecatte, cc @mrhaandi 

The working branch is usually the `master` branch. Could you propose a patch/PR for this branch as well?
The branches named after `rocq` or `coq` are somehow frozen at some given date. Not 100% true and your
proposal might indeed be an exception. However, updates there do not _automatically_ leak into `master`.

Best and thanks for the proposal.

